PR #95, opened 2022-02-23
pds api 127: improved error messaging

al-niessner:
## 🗒️ Summary
Changed all of the AcceptTypeException to return 406 instead of 501.

Improved the message for the exception to include all of the known Accept: content types allowed.

Note, if a Accept: content type is not supported, then the result always comes out as JSON. If the request was 'Accept: text/htlm' then the chosen serializer is JSON.

## ⚙️ Test Data and/or Report
```
$ curl --location --request GET 'http://localhost:8080/products/urn:nasa:pds:izenberg_pdart14_meap:document:ns_ins::1.0' --header 'Accept: text/html'
<html><body><h1>Error Message</h1><h2>From Request</h2><p>/products/urn:nasa:pds:izenberg_pdart14_meap:document:ns_ins::1.0</p><h2>Message</h2><p>The lidvid urn:nasa:pds:izenberg_pdart14_meap:document:ns_ins was not found</p></body></html>
```

```
$ curl --location --request GET 'http://localhost:8080/products/urn:nasa:pds:izenberg_pdart14_meap:document:ns_inst::1.0' --header 'Accept: text/htmls'
{"request":"/products/urn:nasa:pds:izenberg_pdart14_meap:document:ns_inst::1.0","message":"The given application type, text/htmls, is not known by RquestAndResponseContext. Known choices are: application/csv,application/xml,text/html,application/json,*/*,application/pds4+xml,text/csv,text/xml,application/kvp+json,application/pds4+json"}
```

## ♻️ Related Issues

NASA-PDS/registry-api#446

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/business/RequestAndResponseContext.java
+++ b/service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/business/RequestAndResponseContext.java
@@ -161,17 +161,9 @@ public class RequestAndResponseContext
 		}
 		else
 		{
-			String known = "";
-			log.warn("Could not find a matach for application type: " + String.valueOf(this.format));
-			log.warn("   Known types: " + String.valueOf(this.formatters.keySet().size()));
-			for (String key : this.formatters.keySet())
-			{
-				log.warn("      key: " + String.valueOf(key));
-				known = known + key + ",";
-			}
-			known = known.substring(0, known.length()-1);
-			throw new ApplicationTypeException("The given application type, " + String.valueOf(this.format) +
-					                           ", is not known by RquestAndResponseContext. Known choices are: " + known);
+                    String known = String.join(", ", this.formatters.keySet());
+			log.warn("The Accept header value " + String.valueOf(this.format) + " is not supported, supported values are " + known);
+			throw new ApplicationTypeException("The Accept header value " + String.valueOf(this.format) + " is not supported, supported values are " + known);
 		}
 
 		/* if the URL contains fields, then make sure the minimum was included too OR there is maximum set. */
--- a/service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/business/RequestAndResponseContext.java
+++ b/service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/business/RequestAndResponseContext.java
@@ -161,10 +161,17 @@ public class RequestAndResponseContext
 		}
 		else
 		{
+			String known = "";
 			log.warn("Could not find a matach for application type: " + String.valueOf(this.format));
 			log.warn("   Known types: " + String.valueOf(this.formatters.keySet().size()));
-			for (String key : this.formatters.keySet()) log.warn("      key: " + String.valueOf(key));
-			throw new ApplicationTypeException("The given application type, " + String.valueOf(this.format) + ", is not known by RquestAndResponseContext.");
+			for (String key : this.formatters.keySet())
+			{
+				log.warn("      key: " + String.valueOf(key));
+				known = known + key + ",";
+			}
+			known = known.substring(0, known.length()-1);
+			throw new ApplicationTypeException("The given application type, " + String.valueOf(this.format) +
+					                           ", is not known by RquestAndResponseContext. Known choices are: " + known);
 		}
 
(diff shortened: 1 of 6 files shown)


al-niessner:
@tloubrieu-jpl @jordanpadams 

Ready for review.

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/business/RequestAndResponseContext.java
```diff
@@ -161,10 +161,17 @@ private RequestAndResponseContext(
 		}
 		else
 		{
+			String known = "";
 			log.warn("Could not find a matach for application type: " + String.valueOf(this.format));
 			log.warn("   Known types: " + String.valueOf(this.formatters.keySet().size()));
-			for (String key : this.formatters.keySet()) log.warn("      key: " + String.valueOf(key));
-			throw new ApplicationTypeException("The given application type, " + String.valueOf(this.format) + ", is not known by RquestAndResponseContext.");
+			for (String key : this.formatters.keySet())
```
tloubrieu-jpl: Maybe you could use a method like 

String known = String.join(",", this.formatters.keySet());

That would be shorter and I feel like that would be more understandable as well but I did not test it.

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/business/RequestAndResponseContext.java
```diff
@@ -161,10 +161,17 @@ private RequestAndResponseContext(
...
+			{
+				log.warn("      key: " + String.valueOf(key));
```
tloubrieu-jpl: Do we need a warning for this message ? I would think of DEBUG, unless I misunderstood what it does.

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/business/RequestAndResponseContext.java
```diff
@@ -161,10 +161,17 @@ private RequestAndResponseContext(
...
+				known = known + key + ",";
+			}
+			known = known.substring(0, known.length()-1);
+			throw new ApplicationTypeException("The given application type, " + String.valueOf(this.format) +
```
tloubrieu-jpl: I would rephrase the message like:

"The Accept header value " + String.valueOf(this.format) + " is not supported, supported values are " + known

Maybe that sentence does not make sense in the context of the exception but we need a message which makes sense to the user of the API, so that he understands what he need to change in his request.

tloubrieu-jpl (changes requested):
Thanks @al-niessner , that works well. the most critical comment is how the message shows to the user.

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/business/RequestAndResponseContext.java
al-niessner: Need in what way? Will the program work without the warning, yes. When a user says they got a 406 on Saturday will you be able to find it in the log if not posted, no. So define what you mean by need.

Warnings are usually issued when the program detects something fishy (a handled exception) notifying the initiator of the fishy bit that something looks wrong but was handled albeit maybe not how the initiator wanted so they may want to try again -- think or google compiler warnings. Warnings differs from errors in that the program does not crash and burn. Debug on the other hand is way of tracing the program to determine where a bug is hiding or may exist or prove its existance.

For this message when a user mentions days later that a 406 was returned (something fishy but handled) we would have the appropriate level of logging.

If you want it to be debug then simply say so directly and I will change it. One could argue that given errors are being returned to the user the logging is superfluous.

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/business/RequestAndResponseContext.java
al-niessner: resolved

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/business/RequestAndResponseContext.java
al-niessner: resolved

tloubrieu-jpl (approved):
Thanks Al, I will merge it.

PR #99, opened 2022-03-01
Restrict access to registered products

tdddblog:
## 🗒️ Summary
Only return products with archive status = "archived" or "certified"

## ⚙️ Test Data and/or Report
Change archive status of a product by running registry manager, e.g.,
```
registry-manager.bat set-archive-status -status archived -lidvid "urn:nasa:pds:kaguya_grs_spectra:document::1.0"
```
Only "archived" products should be returned by APIs.
NOTE: By default all products are ingested with "staged" status, so unless you change status of some products to "archived", API will not return any products.

## ♻️ Related Issues
https://github.com/NASA-PDS/registry-api/issues/466


tloubrieu-jpl (approved):
To me the pull request is good to be merged. I would like @al-niessner to review it since he's going to maintain and upgrade the code you wrote @tdddblog .

Thanks

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/ElasticSearchRegistrySearchRequestBuilder.java
```diff
@@ -126,89 +122,5 @@ public SearchRequest getSearchCollectionRequest(String queryString, String keywo
...
-        for (Entry<String,List<String>> key : kvps.entrySet())
-        {
-            for (String value : key.getValue())
-            {
-                if (term) find_kvps.should (QueryBuilders.termQuery (key.getKey(), value));
-                else find_kvps.should (QueryBuilders.matchQuery (key.getKey(), value));
-            }
-        }
-        return request;
-    }
```
al-niessner, line 213: Why move this to a new class? The infrastructure for all queries passed through these functions which made it simple to control changes to search options.

Is there a reason putting this functionality in KVPQueryBuilder more efficient?

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/KVPQueryBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+public class KVPQueryBuilder
+{
+    private static final Logger log = LoggerFactory.getLogger(KVPQueryBuilder.class);
+
+    private String esIndex;
+    
+    private List<String> fields;
+    private Map<String,List<String>> kvps;
+    
+    private boolean filterByArchiveStatus = false;
```
al-niessner, line 30: Could have easily added filterByArchiveStatus to the bean RequestAndResponseContext rather than moving all the functionality around and adding yet another bean. Concentrating all of the control context (fields, accept content, etc) makes it simpler to provide consistent behavior and control changes.

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/KVPQueryBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+                boolQuery.must(QueryBuilders.termsQuery(key, values));
+            }
+            else
+            {
+                values.forEach((value) -> 
+                {
+                    boolQuery.must(QueryBuilders.matchQuery(key, value));
+                });
+            }
+        });
```
al-niessner, line 129: This could have been easily added to wordy ElasticSearchRegistrySearchRequestBuilder as a static function like makeStandardSearchRequest() that does this and then replace all the new SearchRequest() with the static function. Doing so means far less change and far less dependence on regression testing -- by regression testing I mean tests that show that bugs that have already been fixed to not regress back into the code.

al-niessner (commented):
If we have robust regression tests and we can show positively that these changes have not reintroduced already fixed bugs, then there is no reason to not simply accept all the changes. However, I am unaware of that level of regression testing capability.

Because our regression testing is limited, I would prefer to see less change and simpler augmentation of the existing infrastructure.

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/ElasticSearchRegistrySearchRequestBuilder.java
tdddblog: There were 7 getQueryFieldsFromKVP() functions with different number of arguments (2 functions had 5 arguments) recursively calling each other. It was very hard to understand what they are doing. I needed to add one more boolean argument and few more functions with the same name. I just replaced all this madness with a new builder class.

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/KVPQueryBuilder.java
tdddblog: The whole point of using builder pattern is to eliminate static methods with 10+ parameters. 
If you look at Elasticsearch APIs or many other modern APIs, they all use builder pattern. For example,
```
new SearchSourceBuilder().query(boolQuery).fetchSource(include, exclude);
```
instead of 
```
StaticSearchSourceBuilder.buildQuery(boolQuery, include, exclude, ...);
```

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/KVPQueryBuilder.java
tdddblog: Context already had methods with 11 parameters and constructor with 13. It would require parameter #14. I've never seen any APIs with that many parameters, but it seems that Java supports 255 :-)
```
RequestAndResponseContext.buildRequestAndResponseContext(this.objectMapper, this.getBaseURL(), lidvid, start, limit, fields, sort, onlySummary, versionSelector, this.presetCriteria, accept);

new RequestAndResponseContext(om, base, null, null, lidvid, start, limit, new ArrayList<String>(), new ArrayList<String>(), false, ProductVersionSelector.ORIGINAL, preset, output_format);
```

service/src/main/java/gov/nasa/pds/api/engineering/elasticsearch/ElasticSearchRegistrySearchRequestBuilder.java
al-niessner: Yes, there are a bunch of he same call because not every caller has all the answers and Java does not allow for default values. So, one ends up with a list of cascading calls that all yet another new default value. I agree that Python like default values would be nicer but this is Java.

By have lowest common denominator builders and spreading out the default values across many places, that leads to a bug and code management madness. So either way, we get to go mad.

My point remains that without robust regression testing this kind of restructuring is dangerous as it ha a high probability of resurrecting fixed bugs.

I am not comfortable with all the change and never will be as many of the bugs fixed in the last two months was removing the large number of small builders that handled default values differently. These changes are just a return to that state. 

@tloubrieu-jpl 
Without a regression check to say that any or none have returned my preference is no more valid than @tdddblog/s so I am not sure of the point of reviewing these changes anymore.

tloubrieu-jpl:
@al-niessner @tdddblog @jordanpadams Per our discussion today, we merge this pull request since the Builder pattern is right way to handle the complexity of the code here.
We accept the risk of regression that is linked to this re-design of the code and will develop better integration test in the near future (postman).

PR #100, opened 2022-03-01
Cloud-tasks-23 : routing of registry-api requests

jimmie:
## 🗒️ Summary
A CloudFront function is introduced (service/terraform/aws/api_url_rewrite.js) which maps the path spec of the URI received at CloudFront (e.g. /api/search/0.4/products) to that needed by the load balancer (/products and the associated x-request-node http header: search/0.4).

A page to locally test the functtion is also provided.

## ♻️ Related Issues
- partially addresses cloud-tasks-19


tloubrieu-jpl (approved):
This pull request is ok for me. Eventually we would need to move the terraform folder into the registry repository because it is not only about the registry api.

PR #110, opened 2022-03-25
Skip ssl certificate verif

tloubrieu-jpl:
## 🗒️ Summary
- add an option to skip SSL certificate verification when connecting to opensearch. This is useful when as a developer/tester we use the docker compose deployment which starts opensearch with ssl certificates which are only valid when requesting opensearch from the internal docker network (as elasticsearch:9200), which is not the case when one develop in an Eclipse IDE deployed on the host (e.g. developer's laptop).
- bump the version to 1.0.0-SNAPSHOT to get ready for v1
- 
## ⚙️ Test Data and/or Report
* Reference to regression test included in code (preferred wherever reasonable)
* Attach test data here + outputs of tests

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
@@ -12,6 +12,7 @@ import org.apache.http.HttpHost;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.CredentialsProvider;
+import org.apache.http.conn.ssl.NoopHostnameVerifier;
 import org.apache.http.ssl.SSLContextBuilder;
 import org.apache.http.ssl.SSLContexts;
 import org.apache.http.conn.ssl.TrustSelfSignedStrategy;
@@ -45,7 +46,7 @@ public class OpenSearchRegistryConnectionImpl implements OpenSearchRegistryConne
 	
 	public OpenSearchRegistryConnectionImpl()
 	{
-	    this(Arrays.asList("localhost:9200"), "registry", "registry-refs", 5, null, null, false);
+	    this(Arrays.asList("localhost:9200"), "registry", "registry-refs", 5, null, null, false, true);
 	}
 	
 	public OpenSearchRegistryConnectionImpl(List<String> hosts, 
@@ -54,7 +55,8 @@ public class OpenSearchRegistryConnectionImpl implements OpenSearchRegistryConne
 			int timeOutSeconds,
 			String username,
 			String password,
-			boolean ssl) {
+			boolean ssl,
+			boolean sslCertificateVerification) {
 		
 		List<HttpHost> httpHosts = new ArrayList<HttpHost>();
 		
@@ -95,6 +97,9 @@ public class OpenSearchRegistryConnectionImpl implements OpenSearchRegistryConne
 						        SSLContext sslContext = sslBld.build();
 	
 						        httpClientBuilder.setSSLContext(sslContext);
+						        if (!sslCertificateVerification) {
+						        	httpClientBuilder.setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE);
+						        }
 			        		}
 				        	
 				            return httpClientBuilder
--- a/service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
@@ -31,6 +31,8 @@ import gov.nasa.pds.api.registry.opensearch.OpenSearchRegistryConnectionImplBuil
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Splitter;
+
 public class OpenSearchRegistryConnectionImpl implements OpenSearchRegistryConnection {
 	
     // key for getting the remotes from cross cluster config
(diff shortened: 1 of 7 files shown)


al-niessner (changes requested):
Is it possible to rebase along main to hide all of the opensearch changes? If not, then let me know and I will spend more time on sifting through the changes.

service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
```diff
@@ -45,7 +46,7 @@
 	
 	public OpenSearchRegistryConnectionImpl()
 	{
-	    this(Arrays.asList("localhost:9200"), "registry", "registry-refs", 5, null, null, false);
+	    this(Arrays.asList("localhost:9200"), "registry", "registry-refs", 5, null, null, false, true);
 	}
 	
 	public OpenSearchRegistryConnectionImpl(List<String> hosts,
```
tdddblog: Constructor has too many parameters. Consider using "builder" design pattern or pass a properties / config / context object.

service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
```diff
@@ -95,6 +97,9 @@ public HttpAsyncClientBuilder customizeHttpClient(
 						        SSLContext sslContext = sslBld.build();
 	
 						        httpClientBuilder.setSSLContext(sslContext);
+						        if (!sslCertificateVerification) {
+						        	httpClientBuilder.setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE);
```
tdddblog, line 99: You only disable host name verification. Don't we also want to trust self-signed certificates?

tdddblog (changes requested):

service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
al-niessner: Actually, should get rid of this constructor altogether -- OpenSearchRegistryConnectionImpl(). The only user of it is TestLidVidUtils which can just hard code this solution.

The only user of other constructor is OpenSearchConfig. If the intent is to say that these values must be supplied and do not have valid or decent defaults (none of them are final which is the other reason to be in the constructor) then the builder pattern is not a good fit here. OpenSearchConfig, being the only caller of the constructor, supplies all values making the current solution sufficient.

tloubrieu-jpl: Actually I remember I create an interface/implementation design while I was struggling with the spring framework to manage the configuration. That might not be useful indeed.

service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
tloubrieu-jpl: @tdddblog I did just what I needed for the api to work with the opensearch within docker network. If the hostname trick works maybe we can just dismiss this pull request.

tloubrieu-jpl:
> Is it possible to rebase along main to hide all of the opensearch changes? If not, then let me know and I will spend more time on sifting through the changes.

@al-niessner i made the rebase

service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
tloubrieu-jpl: @tdddblog I changed the name of the property to be more explicit on the fact that we only disable the hostname verification.

service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
tloubrieu-jpl: @tdddblog @al-niessner I implemented the builder design pattern

al-niessner:
@tloubrieu-jpl 

Sorry but the huge number of files that are changing really make me nervous in a bad way. Is this the only [changeset](https://github.com/NASA-PDS/registry-api/pull/110/commits/a7a5d32cb2a36b11ec363cd7f46f567fd6ccaad3) that is needed for the SSL changes and all others are merging and what not (you are not the committer in the rest)?

al-niessner:
@tloubrieu-jpl 

Actually it looks like these three commits are all that counts:

ssl

Is that correct? If you do not want to do it then I will (I am familiar with doing it), but it would be could to make a patch and start over. With the patch in hand, branch from current HEAD of main, apply patch, make new PR and lets review just the files that are changed. Something is not right with changing 52 files and I am concerned about regressing main to an desirable point in time.

al-niessner (changes requested):
See comments about starting afresh

service/src/main/java/gov/nasa/pds/api/registry/opensearch/OpenSearchRegistryConnectionImpl.java
```diff
@@ -43,25 +44,26 @@
...
-			boolean ssl) {
+			boolean ssl,
+			boolean sslCertificateVerification) {
 		
 		List<HttpHost> httpHosts = new ArrayList<HttpHost>();
 		
-		ElasticSearchRegistryConnectionImpl.log.info("Connection to elastic search");
+		OpenSearchRegistryConnectionImpl.log.info("Connection to open search");
 		for (String host : hosts) {
 			String hostPort[] = host.split(":");
```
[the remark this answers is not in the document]
al-niessner: I would recommend moving to the suggested Splitter class. The idiosyncrasies that this check identifies is exactly the kind that host will produce.

tloubrieu-jpl:
@al-niessner sorry I thought I've done the rebase but it did not work. I did it again and the PR looks better, only showing the changes I actually made.

al-niessner (approved):

Issue #122, opened 2022-04-19
[SECURITY] Upgrade jackson dependencies to remove vulnerability


tloubrieu-jpl:
## 💪 Motivation
Vulnerability in jackson dependencies

## 📖 Additional Details

See automated pull request https://github.com/NASA-PDS/registry-api/pull/119 which can not be merged as is.

See other attempt to upgrade on branch 'jackson-upgrade'. This does not work with the integration test and we had no time to solve that before release 12.1

## ⚖️ Acceptance Criteria
**Given** 
**When I perform** 
**Then I expect**

jordanpadams:
closing per #124 for now. will reopen or issue will pop up again in a future scan

Issue #121, opened 2022-04-12
Fix vulnerabilities raised by sonalift


tloubrieu-jpl:
## 🐛 Describe the bug

See for example https://lift.sonatype.com/results/github.com/NASA-PDS/registry-api/01G0G09HRN3M99XVM8V7BS0KGK

al-niessner:
@tloubrieu-jpl 

Is there an easy way to run these manually?

tloubrieu-jpl:
@al-niessner I have no idea...

tloubrieu-jpl:
Re-opened because of I&T

tloubrieu-jpl:
This ticket need to be solved by:
https://github.com/NASA-PDS/registry-api/issues/219
and/or 
https://github.com/NASA-PDS/registry-api/issues/218

al-niessner:
@jordanpadams @tloubrieu-jpl

The CORS and CRSF are false positives. Yes, the API is open (not authenticated) but nothing is authenticated because everything is GET (stateless). They cannot adversely cause a PUT or change of database state  or any other state as the vulnerability warns (unwanted state changes is the exploit of this vulnerability). Yes, in general, the error is valid but in this context there is not a possible exploit.

If and when we get to the point where you want person A to login to the registry-api to see just mars and jupiter while person B for just venus and mercury then these would exploitable in that A could use B to inadvertently send them venus data. Still, no changes within PDS. Not really seeing any exploits.

jordanpadams:
copy @al-niessner . we plan on eventually implementing some authenticated-access-only capabilities, but I think we will need to address those possible vulnerabilities when the time comes

jordanpadams:
doesn't look like there is anyway we can put on the sonatype Lift issues that these are false positives, so we will just have to use this ticket to track that.

we will wait for @tloubrieu-jpl to call this good or not when he gets back

jordanpadams:
@tloubrieu-jpl can you review this?

jordanpadams:
Closing this as done for now. Will revisit if these vulnerabilities pop back up

PR #124, opened 2022-04-20
issue 121 + 122: fix security vulnerabilities

al-niessner:
## 🗒️ Summary
Update maven package versions.

## ⚙️ Test Data and/or Report
Look at sonatype-lift report

## ♻️ Related Issues
#121 
#122

### DIFF
--- a/model/pom.xml
+++ b/model/pom.xml
@@ -125,12 +125,14 @@
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-web</artifactId>
+      <version>5.3.19</version>
     </dependency>
 
     <!-- https://mvnrepository.com/artifact/org.springframework/spring-context -->
     <dependency>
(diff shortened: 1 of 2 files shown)


al-niessner:
@tloubrieu-jpl @jordanpadams 

LOL. Updated all of the problem packages to the latest and now all of them are showing critical vulnerabilities. The maven repo says they are all vulnerability free though. I suspect this is turtles all the way down. Please advise on what action you want to take.

al-niessner:
@tloubrieu-jpl @jordanpadams 

I keep trolling through the tagged vulnerabilities highlighted by sonatype. Before marking them as ignore or just writing them off, should we design tests in the registry repo to prove/show that we are not using the library inappropriately?

tloubrieu-jpl:
@al-niessner  I don't think we need to develop tests to prove our point that sounds like to much work for that.

Issue #101, opened 2022-03-05
add ops:Tracking_Meta/ops:archive_status to meta section of API response


jordanpadams:
## 💡 Description

Per https://github.com/NASA-PDS/registry-api/issues/466, need this added to API spec and implementation.

tloubrieu-jpl:
@al-niessner @jordanpadams I am wondering why ther tracking_meta and ops:data_files are arrays in the API response, see for example, with request `curl --location --request GET 'http://localhost:8081/products/urn:nasa:pds:insight_rad::2.1' \
--header 'Accept: application/vnd.nasa.pds.pds4+json'`

 ```
      "ops:Data_Files": [
            {
                "ops:file_name": "readme.txt",
                "ops:file_ref": "/data/readme.txt",
                "ops:creation_date": "2020-01-03T17:58:09Z",
                "ops:file_size": "1114",
                "ops:md5_checksum": "192de32c12437c180a9e14d60fe4b89a",
                "ops:mime_type": "text/plain"
            }
        ],
        "ops:Tracking_Meta": [
            {
                "ops:archive_status": "archived"
            }
        ]
```

The label file info is not an array:
```
"ops:Label_File_Info": {
            "ops:file_name": "bundle_insight_rad.xml",
            "ops:file_ref": "/data/bundle_insight_rad.xml",
            "ops:creation_date": "2020-01-15T17:40:30Z",
            "ops:file_size": "6805",
            "ops:md5_checksum": "adfd86bbf2573c37d862e27e08f332db"
        },
```

al-niessner:
@tloubrieu-jpl @jordanpadams 

I did not change the Data_Files or Label_File_Info (that I remember) so they should how they are not because of me but because of how they have always been. I simply cut-n-pasted tracking service from data files. If you want to change them, it is pretty simple but do you want to do it in this ticket or another?

tloubrieu-jpl:
@jordanpadams says that should be a dictionnary

jordanpadams:
should be Data_Files is array, Tracking_Meta is dictionary:

     "ops:Data_Files": [
           {
               "ops:file_name": "readme.txt",
               "ops:file_ref": "/data/readme.txt",
               "ops:creation_date": "2020-01-03T17:58:09Z",
               "ops:file_size": "1114",
               "ops:md5_checksum": "192de32c12437c180a9e14d60fe4b89a",
               "ops:mime_type": "text/plain"
           },
           {
[...]

PR #125, opened 2022-04-20
refactor pds4 metadata to be objects rather than lists

al-niessner:
## 🗒️ Summary
Moved from lists to objects

## ⚙️ Test Data and/or Report
Compiles. The only change was removing a List<x> to x so the compiler finds all the problems.

## ♻️ Related Issues
#101


al-niessner:
@tloubrieu-jpl @jordanpadams 

Changed lists to objects and all works well. All of the integrations tests pass.

al-niessner:
@tloubrieu-jpl @jordanpadams 

fixed

tloubrieu-jpl (approved):
Thanks @al-niessner

Issue #137, opened 2022-05-19
[SECURITY] Log4j vulnerability in lexer


jordanpadams:
## Vulnerability
Need to upgrade log4j to log4j2
* https://github.com/NASA-PDS/registry-api/security/dependabot/2
* https://github.com/NASA-PDS/registry-api/security/dependabot/3
* https://github.com/NASA-PDS/registry-api/security/dependabot/6
* https://github.com/NASA-PDS/registry-api/security/dependabot/7

## Software Version
v1.0.0

tloubrieu-jpl:
Complete the fix with commit 52abb18 

Still need to generate a point build with these updates.

PR #139, opened 2022-05-19
issue 137: [SECURITY] log4j

al-niessner:
## 🗒️ Summary
Purged references to log4j as they are unused

## ⚙️ Test Data and/or Report
See security check after this update

## ♻️ Related Issues
#137


tloubrieu-jpl:
Thanks @al-niessner for your quick fix.

I have a question though, when I run `mvn depedency:tree` is still shows log4j dependency on version 1.2, see:

+- gov.nasa.pds:registry-api-lexer:jar:1.1.0-SNAPSHOT:compile
[INFO] |  +- org.apache.commons:commons-lang3:jar:3.12.0:compile
[INFO] |  +- org.slf4j:slf4j-reload4j:jar:1.7.36:compile
[INFO] |  |  \- ch.qos.reload4j:reload4j:jar:1.2.19:compile
[INFO] |  +- log4j:log4j:jar:1.2.17:compile

I guess that comes from the antlr code generation plugin,

tloubrieu-jpl:
Apparently antlr is not the issue see https://github.com/antlr/antlr4/issues/3428

tloubrieu-jpl:
The log4j 1.2.17 is still in the generated jar:
489884 Sun May 06 13:00:24 PDT 2012 BOOT-INF/lib/log4j-1.2.17.jar

tloubrieu-jpl (approved):
I will merge it anyway since it is certainly not worse than before. We'll see what the security bots say.

tloubrieu-jpl:
:1.1.0-SNAPSHOT is not the right version for the lexer, we are supposed to test with 1.2.0-SNAPSHOT

tloubrieu-jpl:
I re-open the pull request with the `service` module dependency on `lexer` upgraded as it should.

I believe the rest works the same.

Issue #144, opened 2022-06-02
unexpected behavior with q= query


tloubrieu-jpl:
## 🐛 Describe the bug
I had some questions about the API search service:
 
I run this query:
 
[https://pds.nasa.gov/api/search-atm/1.0/products?limit=10&q=(target_name%20eq%20%22Mars%22)](https://urldefense.us/v3/__https:/pds.nasa.gov/api/search-atm/1.0/products?limit=10&q=(target_name*20eq*20*22Mars*22)__;JSUlJQ!!PvBDto6Hs4WbVuu7!cLuBsoFzQjwVd-RL9ckJRuVC60dYqRTlYAexZF_yZBlG-pXynKet1qm2LxlNg_uQ8tcecEFFnA$)
 
but I don't get any results, however, when I run this query:
 
[https://pds.nasa.gov/api/search-atm/1.0/products?limit=10&q=(target_name%20ne%20%22Mars%22)](https://urldefense.us/v3/__https:/pds.nasa.gov/api/search-atm/1.0/products?limit=10&q=(target_name*20ne*20*22Mars*22)__;JSUlJQ!!PvBDto6Hs4WbVuu7!cLuBsoFzQjwVd-RL9ckJRuVC60dYqRTlYAexZF_yZBlG-pXynKet1qm2LxlNg_uQ8tcRNh2iwg$)
 
I do get Mars results.
 
Can you please tell me what I'm doing wrong?

tloubrieu-jpl:
Bug initially reported by Irma Trejo from atmosphere node.

al-niessner:
@tloubrieu-jpl 

From a [change](https://github.com/NASA-PDS/registry-api/commit/900fd0e99a46fab414da25616e3b226aff4bfec9) @jordanpadams  may have approved it seems that eq has change to like.

tloubrieu-jpl:
This ticket is closed as a duplicate of https://github.com/NASA-PDS/registry-api/issues/134

Issue #134, opened 2022-05-16
wildcard search in query parameter returning odd results


tloubrieu-jpl:
## 🐛 Describe the bug

I am not sure about the implementation of wildcard in the API, however I submit a bug the result of this request  does not make sense to me:

```
curl --location --request GET 'http://localhost:8080/products?limit=10&q=title eq "InSight RAD*"&fields=title' \
--header 'Accept: text/csv'
```

Returns:

```
title
"InSight RAD Derived Data Collection"
"InSight RAD Raw Data Collection"
"InSight HP3 Radiometer Experiment Derived Product:hp3_rad_der_00014_20181211_073042"
"InSight HP3 Radiometer Experiment Derived Product:hp3_rad_der_00122_20190401_123217"
"InSight HP3 Radiometer Experiment Derived Product:hp3_rad_der_00213_20190703_052044"
"InSight HP3 Radiometer Experiment Derived Product:hp3_rad_der_00305_20191006_053040"
"InSight HP3 Radiometer Experiment Derived Product:hp3_rad_der_00390_20200101_120222"
"InSight HP3 Radiometer Experiment Derived Product:hp3_rad_der_00478_20200401_121608"
"InSight HP3 Radiometer Experiment Derived Product:hp3_rad_der_00567_20200701_125545"
"InSight HP3 Radiometer Experiment Raw Product:hp3_rad_raw_00004_20181130_085325"
```

## 🕵️ Expected behavior

Return only products which title start with "InSight RAD".

## 📚 Version of Software Used

version 1.0

## 🖥 System Info

---

al-niessner:
@tloubrieu-jpl 

From a [change](https://github.com/NASA-PDS/registry-api/commit/900fd0e99a46fab414da25616e3b226aff4bfec9) @jordanpadams  may have approved it seems that eq has change to like.

jordanpadams:
@al-niessner @tloubrieu-jpl hmmm. apologies there. I think I may have described what I wanted incorrectly. 😥 

The way I imagine "eq" functionality to work is it matches "eq" as you would think (e.g. ==) but also supports wildcards. So not quite "like" but does still evaluate wildcards. For instance:

`title eq "InSight RAD*"`:
| title | matches? |
|-------|----------|
| InSight RAD 123 | X |
| InSight RAD | X |
| foobar InSight RAD | X **???** |
| insight rad 123 | X |
| InSight foobar RAD | |

[...]

PR #158, opened 2022-07-11
changed string searching

al-niessner:
## 🗒️ Summary

Re-enabled the unit testing for the queries that use the local language.

Seems that wildcards is not well implemented in the later elasticsearch and opensearch implementations. The most recent documentation for either has changed significantly that if searhing for strings, then they offer a short globish like set of patterns. Moved from using wildcards to this new string and it seems to be working.

The eq operator had some problems because it was using should instead of filter for the or condition. SHould returns everything with a score value to represent the closes match from the furthest. Changed over to filter and the or condition is now behaving as expected.

## ⚙️ Test Data and/or Report

Unittests pass:

```
$ curl --location --request GET 'http://localhost:8080/gid/any?limit=10&fields=title&q=title%20like%20%22Insight%2BRAD%22' --header 'Accept: text/csv'
title
"InSight RAD Calibrated Data Collection"
"InSight RAD Derived Data Collection"
"InSight RAD Raw Data Collection"
```

```
$ curl --location --request GET 'http://localhost:8080/gid/any?limit=10&fields=title&q=title%20eq%20%22InSight%20RAD%2A%22' --header 'Accept: text/csv'

$
```

## ♻️ Related Issues
#134 
#144

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/model/Antlr4SearchListener.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/model/Antlr4SearchListener.java
@@ -13,9 +13,9 @@ import java.util.List;
 
 import org.antlr.v4.runtime.misc.ParseCancellationException;
 import org.opensearch.index.query.BoolQueryBuilder;
-import org.opensearch.index.query.MatchQueryBuilder;
-import org.opensearch.index.query.WildcardQueryBuilder;
 import org.opensearch.index.query.RangeQueryBuilder;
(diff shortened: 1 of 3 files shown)


al-niessner:
@tloubrieu-jpl @jordanpadams 

Needs review. When done, will move it to release 1.0.1 as well.

tloubrieu-jpl:
Hi @al-niessner ,

That works except for the like that I have not been able to make work. It is not part of this pull request, right ?

tloubrieu-jpl (approved):
Good, I will create a ticket or the `like`

PR #183, opened 2022-09-22
Update pom versions to inherit from parent

### LINKED ISSUE #182: Sub-project should inherit parent project version in order to be able to automate versioning using mvn commands
## 🐛 Describe the bug

Whenever the Roundup Action executes `mvn versions:set` it affects only the repository's `pom.xml`. However, [a certain repository whose name will not be spoken](https://github.com/NASA-PDS/registry-api) is a "multi-module" repository whose `pom.xml` contains
```xml
<modules>
    <module>model</module>
    <module>lexer</module>
    <module>service</module>
</modules>
```
But `versions:set` doesn't update the `<version>` specification in child `pom.xml` files.

Apparently instead of just

    mvn -DnewVersion=1.2.3 versions:set

Rounup needs to do

    mvn -DnewVersion=1.2.3 -DprocessAllModules=true -DartifcatId=* -DgroupId=* versions:set

## 🕵️ Expected behavior

In `registry-api`, all `pom.xml` files should be updated after `versions:set`.

## 📚 Version of Software Used

`roundup-action@stable`

---

jordanpadams:
To enable consistency across modules and automation of build.

Resolves #182


jimmie (approved):
Nothing short of brilliant.

PR #209, opened 2022-12-14
registry-api-208

jimmie:
## 🗒️ Summary
registry-api-208 : 
     replace multiple slashies w/ a single
     strip out minor version numbers in full requests

Went ahead w/ the strategy of replacing all multiple slashies in the post-host URI.


tloubrieu-jpl (approved):
Thanks Jimmie

PR #221, opened 2022-12-23
Make application.properties.aws file agree w/ application.properties

jimmie:
## 🗒️ Summary
Need to establish parity between application.properties.aws and application.properties by updating the former to reflect recent changes/additions in the latter. Some differences need to remain (e.g. openSearch host, username, password remain empty to invoke reading from environment, port number remains at 80)

### DIFF
--- a/service/src/main/resources/application.properties.aws
+++ b/service/src/main/resources/application.properties.aws
@@ -1,15 +1,15 @@
 springfox.documentation.swagger.v2.path=/api-docs
+server.contextPath=/
+server.port=80
+server.forward-headers-strategy=framework
+
 spring.jackson.serialization.WRITE_DATES_AS_TIMESTAMPS=false
+springfox.documentation.enabled=false
 
 debug=true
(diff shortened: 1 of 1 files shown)


jimmie:
As noted in the original issue:

Note that this file (along with the registry-api jar file) is read in at the time the docker image is built so a new build of the registry-api artifacts is not necessary to support these changes.

jordanpadams:
@jimmie just to verify, the following is then being captured elsewhere right now?
```
server.contextPath=/
server.port=80
server.use-forward-headers=true
```

jordanpadams:
@jimmie ☝️

jimmie:
> @jimmie just to verify, the following is then being captured elsewhere right now?
> 
> ```
> server.contextPath=/
> server.port=80
> server.use-forward-headers=true
> ```

They are still there, just shifted up to be in the same location as the other properties file.

jordanpadams (approved):

PR #233, opened 2023-01-18
Provenance script tweaks

alexdunnjpl:
## 🗒️ Summary
This started out as an attempt to understand the script and figure out why it ostensibly wasn't working on my local dev instance (it was - the API version I was using was old).

Cliffnotes:
- argument parsing is moved to the main block to allow `run()` to be called from other contexts
- tweaked logging and changed default level to INFO (otherwise, a successful run yields no output, which is confusing)
- simplified some status checks that were just throwing on non-200
- renamed some labels for clarity 
- removed extraneous no-op statements
- Python-consistent styling (whitespace, line-breaks).  I won't say PEP because it wasn't a comprehensive lint.

I've left the commits granular in case some of these changes aren't desirable (kick them back to me and I'll fix)

### DIFF
--- a/support/provenance.py
+++ b/support/provenance.py
@@ -18,7 +18,8 @@ HOST = collections.namedtuple('HOST', ['nodes', 'password', 'url', 'username',
                                        'verify'])
 
 
-def _log_level(input: str) -> int:
+def parse_log_level(input: str) -> int:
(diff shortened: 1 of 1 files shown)


jimmie (approved):
Nice changes.

Issue #234, opened 2023-01-19
registry-api does not respect VID when a LIDVID is used as an id, instead returns latest version


alexdunnjpl:
## 🐛 Describe the bug
Given multiple ingested versions of a product, requesting `<someclass>/<someLIDVID>` returns the document for the most-recent version of the product matching the LID, irrespective of which VID is specified

## 📜 To Reproduce
1. Deploy docker-composed registry `int-registry-batch-loader` profile (which includes `urn:nasa:pds:insight_rad:data_derived::7.0`
2. Run `harvest` against the latest version of `urn:nasa:pds:insight_rad` bundle
3. Run `registry-mgr` against the LIDVID of the bundle or `data_derived` collection
4. Run `curl --get 'http://localhost:8080/collections/urn:nasa:pds:insight_rad:data_derived::7.0' | json_pp`, observe v14.0 metadata returned

## 🕵️ Expected behavior
Expect return of requested version, not latest

## 📚 Version of Software Used
Docker image `nasapds/registry-api-service:1.1.11`
registry-api version which includes commit 252bfbe0 and its dependencies

alexdunnjpl:
This appears to be caused by [the default value of URIParameters.selector](https://github.com/NASA-PDS/registry-api/blob/main/service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java#L45) which is always `ProductVersionSelector.LATEST`.

My gut feeling is that the value of `selector` should depend on the value of `lidvid` and _probably_ be set to `ProductVersionSelector.TYPED` if the supplied lidvid is actually a LIDVID and not just a LID, but I'm not yet sure of what side-effects this might cause.

jimmie:
arose from registry-api #223

al-niessner:
@alexdunnjpl @jimmie 

No, this VID problem is a sorting problem as stated in yet another ticket that I cannot find but I think @alexdunnjpl implemented a more complex set of classes to be able to sort to fix it. It has nothing to do with URIParameters.

alexdunnjpl:
@al-niessner [This is probably the ticket in question (hidden because it's closed)](https://github.com/NASA-PDS/registry-api/issues/224)

That was a sorting issue, but if I understand what `ProductVersionSelector` is supposed to do, asking the API for `urn:nasa:pds:insight_rad:data_derived::7.0` should result in a value of `TYPED` because the user typed the version they needed?

[...]

PR #235, opened 2023-01-19
Add classes for LID/VID/LIDVID and fix #224

### LINKED ISSUE #224: registry-api does not return latest version of product metadata when multiple versions are harvested
## 🐛 Describe the bug
Given that v7.0 of the example product is harvested and set to archived status, if v14.0 is harvested and ingested:
- Requesting the LID returns v7.0 metadata
- Requesting the v14.0 LIDVID succeeds, but returns v7.0 metadata.

## 📜 To Reproduce
1. Deploy docker-composed registry `int-registry-batch-loader` profile (which includes `urn:nasa:pds:insight_rad:data_derived::7.0`
2. Run `harvest` against the latest version of `urn:nasa:pds:insight_rad` bundle
3. Run `registry-mgr` against the LIDVID of the bundle or `data_derived` collection
4. Run `curl --get 'http://localhost:8080/collections/urn:nasa:pds:insight_rad:data_derived' | json_pp`, observe v7.0 metadata returned
5. Run `curl --get 'http://localhost:8080/collections/urn:nasa:pds:insight_rad:data_derived::14.0' | json_pp` (i.e latest version LIDVID), observe v7.0 metadata returned

## 🕵️ Expected behavior
Expect return of v14.0 metadata

## 📚 Version of Software Used
Docker image `nasapds/registry-api-service:1.1.11`

alexdunnjpl:
## 🗒️ Summary
Introduces LID/VID/LIDVID as first-class objects, including comparison operations.

## ⚙️ Test Data and/or Report
Unit tests added for new classes, manually-tested test cases for #224 

## ♻️ Related Issues
fixes #224
causes #234 as a side-effect of fixing #224 (but I'll get to that)

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/model/PdsLid.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/model/PdsLid.java
@@ -1,0 +1,36 @@
+package gov.nasa.pds.api.registry.model;
+
+public class PdsLid {
+    private String value;
+
+    public PdsLid(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return this.value;
+    }
+    @Override
+    public String toString() {
+        return this.value;
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) {
+            return true;
+        }
+
+        if (!(other instanceof PdsLid)) {
+            return false;
+        }
+
+        PdsLid otherAsLid = (PdsLid) other;
+        return this.value.equals(otherAsLid.getValue());
+    }
+
+    @Override
+    public int hashCode() {
+        return this.value.hashCode();
+    }
+}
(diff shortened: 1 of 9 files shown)


alexdunnjpl:
Sorry for all the whitespace changes - do we have an autoformat configuration for our projects kicking around?  I'm just using the IDEA defaults currently.

service/src/main/java/gov/nasa/pds/api/registry/model/PdsLid.java
```diff
@@ -0,0 +1,36 @@
+package gov.nasa.pds.api.registry.model;
```
nutjob4life, line 1: Kind of surprised we don't already have a "data class" like this, perhaps in `pds4-jparser`. Oh well, better late than never!

nutjob4life (approved):
LGTM

service/src/main/java/gov/nasa/pds/api/registry/model/PdsLid.java
alexdunnjpl: We may do, but if so it's not in use in _this_ repo.  Is it worth me following that up, or am I good to continue with this for this repo, at least?

alexdunnjpl: You know what?  After looking at it I think that module might be _slight_ overkill for the task of sorting VIDs (actually I don't even think it's capable of that, but it *is* very imposing and impressive)

PR #236, opened 2023-01-19
Add additional lidvid classes

### LINKED ISSUE #234: registry-api does not respect VID when a LIDVID is used as an id, instead returns latest version

alexdunnjpl:
## 🗒️ Summary
Implements a `PdsProductIdentifier` parent class of `PdsLid` and `PdsLidVid` and sinks its teeth into the codebase (i.e. changes interfaces of a large quantity of functions to use the new classes instead of `String`s).  This relates to #234 but is much bigger than a simple fix for that one ticket.

I don't expect anyone to read through the bulk of it, but I'd like to get feedback on the implementation of the `PdsProductIdentifier` class, and the general position that LID/LIDVID strings should be converted to/from `PdsProductIdentifier` objects as close as possible to the API's/Java's interface with external services.

If anyone _does_ hate themselves enough to look closely, bear in mind that it's a work in progress and the expansion of mandatory `PdsProductIdentifier`-ness necessitates a bunch of temporary code at the incrementally-expanding interfaces (hence all the streaming collections back and forth between `Strings` and `PdsProductIdentifiers`).

The job's only half-done, but better that I get feedback now than after spending a bunch more time if there are problems with the general approach.

Resolves #234

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/model/RefLogicCollection.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/model/RefLogicCollection.java
@@ -9,8 +9,11 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import gov.nasa.pds.api.registry.model.identifiers.LidVidUtils;
+import gov.nasa.pds.api.registry.model.identifiers.PdsLid;
+import gov.nasa.pds.api.registry.model.identifiers.PdsLidVid;
 import gov.nasa.pds.api.registry.model.identifiers.PdsProductIdentifier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -83,9 +86,10 @@ class RefLogicCollection extends RefLogicAny implements ReferencingLogic
     static Pagination<String> parents (ControlContext control, ProductVersionSelector selection,  LidvidsContext uid)
     		throws IOException, LidVidNotFoundException
     {
+//        TODO: Fully convert this function's internals (and eventually, interface) to use PdsProductIdentifier classes instead of strings
     	List<String> keys = Arrays.asList("ref_lid_collection", "ref_lid_collection_secondary",
     		                              "ref_lidvid_collection", "ref_lidvid_collection_secondary");
-    	List<String> sortedLids;
+    	List<String> sortedLidStrings;
     	Set<String> lids = new HashSet<String>();
     	PdsProductIdentifier productIdentifier = PdsProductIdentifier.fromString(uid.getLidVid());
         PaginationLidvidBuilder bundleLidvids = new PaginationLidvidBuilder(uid);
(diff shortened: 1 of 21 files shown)


nutjob4life (commented):
Modern! This looks fine to me. Do we want `hashCode`?

service/src/main/java/gov/nasa/pds/api/registry/model/identifiers/LidVidUtils.java
```diff
@@ -120,33 +109,36 @@ public static List<String> getAllLidVidsByLids(
     }
 
     public static String resolve (
-    		String identifier,
+    		String _identifier,
```
ramesh-maddegoda, line 112: Variable names should not start with underscore _ or dollar sign $ characters, even though both are allowed.
https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html#:~:text=Variable%20names%20should%20not%20start,should%20be%20short%20yet%20meaningful.
https://google.github.io/styleguide/javaguide.html

alexdunnjpl: Roger that, will fix - this is an artifact of me not having yet converted the function's interface to use the classes.

service/src/main/java/gov/nasa/pds/api/registry/model/RefLogicCollection.java
```diff
@@ -81,32 +86,39 @@ private static Pagination<String> childrenLatest (ControlContext control, Lidvid
...
         					control.getConnection().getRegistryIndex())))
         	{
         		lids.addAll(bundleLidvids.convert(kvp.get("lid")));
         	}
         }
-        sortedLids = new ArrayList<String>(lids);
-        Collections.sort(sortedLids);
+        sortedLidStrings = new ArrayList<String>(lids);
+        Collections.sort(sortedLidStrings);  // TODO: Implement comparison for PdsLids (only with other PdsLids)
+		List<PdsProductIdentifier> sortedLids = sortedLidStrings.stream().map(PdsLid::fromString).collect(Collectors.toList());
```
ramesh-maddegoda: Unnecessary indent here (line 111, before List)?

alexdunnjpl:
~Breaks (at least) `http://localhost:8080/classes/collections/` endpoint, as-is.~

Fixed in [992aa43](https://github.com/NASA-PDS/registry-api/pull/236/commits/992aa4349851e272d31d94f6099022d38c271490)

alexdunnjpl:
@al-niessner @tloubrieu-jpl @nutjob4life @ramesh-maddegoda 

Can I get a second pair of eyes on [this logging commit](https://github.com/NASA-PDS/registry-api/pull/236/commits/22a0970d2f23046ba07842c21da2076cebab16fc)?  It's as not-garbage as I can think to make it but maybe there's an obviously better way.

nutjob4life (approved):
Sure 👍

jordanpadams:
@alexdunnjpl can we update this PR Summary with the ticket this is related to?

alexdunnjpl:
@jordanpadams done

alexdunnjpl:
@al-niessner courtesy ping so you don't have to look for the PR.  Just the last 3 commits.

resolve() case`LATEST` now has an explanatory comment (which I hope is accurate), and behaves identically to case `SPECIFIC` (which I just realised I should change back to case `TYPED` - will do that shortly).

Existing `RequestConstructionContext`'s `getLIDVID()` is changed to `getProductIdentifierString()`, and the identifier member is swapped out for a `PdsProductIdentifier` in the implementation internals.

This allows for correction of `RequestAndResponseContext.justLatest()` to respect the fact that if a full LIDVID is the product identifier associated with the request, then the query should *not* filter to only the latest products because the version is specified (and may not be a latest-version product)

service/src/main/java/gov/nasa/pds/api/registry/model/identifiers/LidVidUtils.java
alexdunnjpl: resolved in 9c6653c19c98f0b268887f11f4e8dc2f6736b268

PR #246, opened 2023-02-16
bump opensearch page size 10 -> 500

### LINKED ISSUE #231: /classes/collections/<lidvid>/members (and deprecated equivalent) hangs
## 🐛 Describe the bug
The endpoints /classes/collections/<lidvid>/members and /collections/<lidvid>/products hang.

## 📜 To Reproduce
1. curl https://pds.nasa.gov/api/search/1/collections/urn:nasa:pds:orex.ovirs:data_calibrated/products
2. no results are returned
3. request eventually times out w/ a gateway error

## 🕵️ Expected behavior
Expect a response containing the products contained by the collection to be returned, in a timely manner.

## 📚 Version of Software Used
registry-api 1.1.12

alexdunnjpl:
## 🗒️ Summary
Bumps OpenSearch default page size to 500
Also rearranges a couple of calls (no functional impact)

## ⚙️ Test Data and/or Report
See GH tests

## ♻️ Related Issues
fixes #231

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/model/RefLogicCollection.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/model/RefLogicCollection.java
@@ -11,20 +11,24 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import gov.nasa.pds.api.registry.model.identifiers.LidVidUtils;
-import gov.nasa.pds.api.registry.model.identifiers.PdsLid;
-import gov.nasa.pds.api.registry.model.identifiers.PdsLidVid;
-import gov.nasa.pds.api.registry.model.identifiers.PdsProductIdentifier;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.google.errorprone.annotations.Immutable;
 
 import gov.nasa.pds.api.registry.ControlContext;
 import gov.nasa.pds.api.registry.GroupConstraint;
 import gov.nasa.pds.api.registry.LidvidsContext;
 import gov.nasa.pds.api.registry.ReferencingLogic;
+import gov.nasa.pds.api.registry.RequestConstructionContext;
+import gov.nasa.pds.api.registry.RequestBuildContext;
 import gov.nasa.pds.api.registry.UserContext;
+import gov.nasa.pds.api.registry.model.identifiers.LidVidUtils;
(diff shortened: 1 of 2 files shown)


service/src/main/java/gov/nasa/pds/api/registry/model/RefLogicCollection.java
```diff
@@ -11,20 +11,17 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import gov.nasa.pds.api.registry.*;
```
nutjob4life: Wildcard imports 🤮

service/src/main/java/gov/nasa/pds/api/registry/search/HitIterator.java
```diff
@@ -12,12 +12,12 @@
 
 public class HitIterator implements Iterable<Map<String,Object>>,Iterator<Map<String,Object>>
 {
-	private int size=10; // define size to use here to prevent page skipping if opensearch default size ever changes
+	private int size=500; // define size to use here to prevent page skipping if opensearch default size ever changes
```
nutjob4life, line 15: So this is all it takes? Just bump this up? If OpenSearch changes again will we need to change this again?

nutjob4life (approved):

service/src/main/java/gov/nasa/pds/api/registry/model/RefLogicCollection.java
alexdunnjpl: Whoops, that was some "helpful" IDE intervention that I should've caught.  Will fix (tyvm for the catch!)

service/src/main/java/gov/nasa/pds/api/registry/search/HitIterator.java
alexdunnjpl: Negative(ish) - this is primarily a factor of data volume.  Given `HitIterator.size=n` and some OpenSearch internal paging config value `m` (currently set to 500 and hidden deep enough that I doubt it's likely to change unless something breaks), HitIterator will make a somewhat-expensive request for every `n*m` products it needs to fetch.

If the thrust is "Hm, seems like this should be a config value bro" then yes, yes it should (and should be extracted at $laterDate), but I'm changing it here for expediency's sake.

nutjob4life: Roger that!

Issue #250, opened 2023-02-16
API does not appear to be applying LIDVID-sort by default


alexdunnjpl:
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

[When I used a test script to validate pagination correctness](https://github.com/NASA-PDS/registry-api/issues/247#issuecomment-1433584682), results indicated that sorting did not appear to be correctly applied.

If true, this completely breaks ability to query across multiple pages of data as some LIDVIDs will be duplicated while others will be missing.

### 🕵️ Expected behavior

Given values of `start` and `limit`, I expect them to provide repeatable and consistent slices of data even when an explicit sort queryparam is not provided

### 📜 To Reproduce

TBD - need to fix #240 first and then validate that the test script isn't what's borked.

alexdunnjpl:
Have confirmed this behaviour with script.

Testing starts [0,3] with limits [1,3], the following hashed lidvids are obtained:

```
['030037', '      ', '      ', '      ', '      ', '      ']
['1d58f1', '030037', '      ', '      ', '      ', '      ']
['1d58f1', '86c734', '030037', '      ', '      ', '      ']
['      ', '1d58f1', '      ', '      ', '      ', '      ']
['      ', '1d58f1', '86c734', '      ', '      ', '      ']
['      ', '1d58f1', '86c734', '66fdfc', '      ', '      ']
['      ', '      ', '86c734', '      ', '      ', '      ']
['      ', '      ', '86c734', '66fdfc', '      ', '      ']
['      ', '      ', '96d8ce', '86c734', '66fdfc', '      ']
['      ', '      ', '      ', '66fdfc', '      ', '      ']
['      ', '      ', '      ', '96d8ce', '66fdfc', '      ']
['      ', '      ', '      ', '96d8ce', '756d24', '66fdfc']
```

alexdunnjpl:
More test data from start 0, limits [1,11]

```
96d8ce - urn:nasa:pds:mars2020_supercam:data_calibrated_audio:scam_0012_0668002890_477_ca1_scam15210_maaz_________________01p::1.1
a7492c - urn:nasa:pds:mars2020_supercam:data_calibrated_audio:scam_0037_0670224605_709_ca0_scam01037_hedgehog_____________04p::1.1
1d58f1 - urn:nasa:pds:mars2020_supercam:data_calibrated_audio:scam_0004_0667311609_463_ca2_scam15104_atm__________________00p::1.1
[...]

PR #251, opened 2023-02-17
Fix pagination bug

### LINKED ISSUE #240: Pagination not working as expected with /collections/{identifier}/products
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

@jordanpadams thought the [deep-archive](https://github.com/NASA-PDS/deep-archive/) was paginating incorrectly. I know it worked at one time.

The [documentation](https://nasa-pds.github.io/pds-api/specifications/search-v1.0.0-redoc.html#operation/products-of-a-collection) says the `limit` is a cardinal, telling _how many_ to return. But he suspected that it might now be an ordinal, telling the _end index_ of the row to return.

I [tested both extensively in a comment on a separate issues](https://github.com/NASA-PDS/deep-archive/issues/140#issuecomment-1404269589) and found that it's neither. It's sort of just broken.

### 🕵️ Expected behavior

I expected to paginate by using `start` and `limit` parameters, preferably with `limit` telling _how many_ to return.

### 📜 To Reproduce

```bash
See https://github.com/NASA-PDS/deep-archive/issues/140#issuecomment-1404269589
```

### 🖥 Environment Info

```bash
- https://pds.nasa.gov/api/search/1.0 as of 2023-01-26
```

- https://pds.nasa.gov/api/search/1.0 as of 2023-01-26

_No response_

### 🦄 Related requirements

N/A

### ⚙️ Engineering Details

May be tightly coupled with https://github.com/NASA-PDS/registry-api/issues/231

alexdunnjpl:
## 🗒️ Summary
Fixes a bug in `PaginationLidvidBuilder` which resulted in incorrect slicing, as well as crashes when provided `start` parameter was greater than provided `limit` 

## ⚙️ Test Data and/or Report
Added comprehensive unit tests for `PaginationLidvidBuilder.addAll()`

## ♻️ Related Issues
Fixes #240 

#250 outstanding

### DIFF
--- a/service/src/test/java/gov/nasa/pds/api/registry/model/PaginationLidvidBuilderTest.java
+++ b/service/src/test/java/gov/nasa/pds/api/registry/model/PaginationLidvidBuilderTest.java
@@ -1,0 +1,273 @@
+package gov.nasa.pds.api.registry.model;
+
+import gov.nasa.pds.api.registry.LidvidsContext;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class PaginationLidvidBuilderTest {
+
+    private final List<String> simpleInput = Arrays.asList("1", "2", "3", "4", "5");
+    private final List<List<String>> complexInput = Arrays.asList(
+            Arrays.asList("1", "2", "3"),
+            Arrays.asList("4", "5", "6"),
+            Arrays.asList("7", "8", "9")
+    );
+
+    @Test
+    public void testSimpleAddTrimNothing() {
+        PaginationLidvidBuilder pageBuilder = new PaginationLidvidBuilder(new LidvidsContextStub(0, 5));
+        pageBuilder.addAll(simpleInput);
+        Assert.assertArrayEquals(new String[]{"1", "2", "3", "4", "5"}, pageBuilder.page().toArray());
+    }
+
+    @Test
+    public void testSimpleAddTrimEverything() {
+        PaginationLidvidBuilder pageBuilder = new PaginationLidvidBuilder(new LidvidsContextStub(0, 0));
+        pageBuilder.addAll(simpleInput);
+        Assert.assertArrayEquals(new String[]{}, pageBuilder.page().toArray());
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


service/src/test/java/gov/nasa/pds/api/registry/model/PaginationLidvidBuilderTest.java
```diff
@@ -0,0 +1,273 @@
...
+        for (List<String> input : complexInput) {
+            pageBuilder.addAll(input);
+        }
+        Assert.assertArrayEquals(new String[]{"8"}, pageBuilder.page().toArray());
+    }
+
+
+    @Test
+    public void testComplexAddTrimHeadAndTailSpanning() {
+        PaginationLidvidBuilder pageBuilder = new PaginationLidvidBuilder(new LidvidsContextStub(2, 5));
```
nutjob4life, line 236: Try any negatives like `LidvidsContextStub(2, -5)` or `LidvidsContextStub(-2, -5)`?

nutjob4life (approved):
👍

service/src/test/java/gov/nasa/pds/api/registry/model/PaginationLidvidBuilderTest.java
alexdunnjpl: Seemed defensible (albeit naive) to assume that a passed `LidvidsContext` would be valid.  If that validation isn't already baked into the existing `LidvidsContext` implementations I'll see what I can do there - _might_ be trivial but I need to verify that start/limit aren't used in any context where negatives are (say) expected, but handled after instantiation.

Plus I'll need to see how to throw in a way that will return a 422, which may be more trouble than is worth doing right this second (but I'll add it to the backlog).

tloubrieu-jpl: Ok to test raise an exception when limit or start is negative. Ok also to keep 422 error for later. 400 with a dedicated message sounds more than enough to me for now. Thanks

alexdunnjpl:
Merging, and will open a new issue for lower bounds (as it's out of scope for 240) and fix asap.

Issue #232, opened 2023-01-17
Move scheduled execution of provenance.py to an AWS-based solution


jimmie:
## 💡 Description
The provenance.py script will soon be deployed on pdscloud-prod1 in pds4's crontab. We should consider employing this script in an AWS-based manner, such as AWS Batch.

This also provides the means by which we can use something like Secrets Manager to maintain the Opensearch credentials.

jimmie:
After discussions during the SA office hour, we propose the approach of AWS EventBridge Scheduler (obviously for scheduling) driving the invocation of the provenance as an ECS (Fargate) standalone task. Credentials will be injected into the ECS container as environment variables and other runtime parameters from the Parameter Store.

Greatest amount of work is to create the container image that runs the script.

jimmie:
Solution design is documented here: https://wiki.jpl.nasa.gov/pages/viewpage.action?spaceKey=PDSEN&title=Registry+Provenance+Execution

jordanpadams:
closed per https://github.com/NASA-PDS/registry-api/pull/253

Issue #160, opened 2022-07-14
apply bug fix on stable release


tloubrieu-jpl:
## 💡 Description

Apply the pull request changes from https://github.com/NASA-PDS/registry-api/pull/158 on branch release/1.0.1

al-niessner:
@tloubrieu-jpl @jordanpadams 

# DANGER!!

Seems that branch release/1.0.1 no longer exists. Did somebody do a push force?

If you go to #147 then click on the [release/1.0.1](https://github.com/NASA-PDS/registry-api/tree/release/1.0.1) that existed 12 days ago you will get a 404 page.

I have this fix sitting on branch issue_134_r101 (should have named it issue 160 but oh well) to make a PR and merge but do not know which branch to use now that release 1.0.1 is MIA.

tloubrieu-jpl:
Hi @al-niessner ,

I still have the branch on my local repository. I wil repush it with a .bkp suffix not to override whatever happened to the original branch.

tloubrieu-jpl:
I pushed the branch https://github.com/NASA-PDS/registry-api/tree/release/1.0.1.bkp which should be the same as the original release/1.0.1

PR #253, opened 2023-02-22
Scheduled execution of provenance script

jimmie:
## 🗒️ Summary
Provide scheduled execution of provenance script in AWS, employing event-bridge scheduler initiating stand-alone Fargate execution. Commits include:

- terraform scripts to deploy schedule, Fargate task and service parameters
- python driver script which wraps provenance.py to adapt it to Fargate execution
- docker file to create ECR image

Also creates a provenance sub-directory beneath support to keep all relevant files in an isolated location.

## ♻️ Related Issues
[registry-api #232 ](https://github.com/NASA-PDS/registry-api/issues/232)
[registry #160](https://github.com/NASA-PDS/registry/issues/160)

### DIFF
--- a/support/provenance/provenance.py
+++ b/support/provenance/provenance.py
@@ -1,5 +1,46 @@
 #! /usr/bin/env python3
 
+# Copyright © 2023, California Institute of Technology ("Caltech").
+# U.S. Government sponsorship acknowledged.
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# • Redistributions of source code must retain the above copyright notice,
+#   this list of conditions and the following disclaimer.
+# • Redistributions must reproduce the above copyright notice, this list of
+#   conditions and the following disclaimer in the documentation and/or other
+#   materials provided with the distribution.
+# • Neither the name of Caltech nor its operating division, the Jet Propulsion
+#   Laboratory, nor the names of its contributors may be used to endorse or
+#   promote products derived from this software without specific prior written
+#   permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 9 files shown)


support/provenance/provenance.py
```diff
@@ -92,6 +133,9 @@ def get_historic(provenance: {str: str}, reset: bool) -> {str: str}:  # TODO: po
 
     log.info(
         f'found {len(history)} products needing update of a {count} full history of {len(provenance)} total products')
+    if log.isEnabledFor(logging.DEBUG):
+        for lidvid in history.keys():
+            log.debug(f'{lidvid}')
```
tloubrieu-jpl, line 138: We might be able to have a single line for that, if we don't need a single debug message for each lidvid

log.debug(', '.join(history.keys()))

or 

log.debug('\n'.join(history.keys()))

i have not tested that though

support/provenance/provenance_driver.py
```diff
@@ -0,0 +1,90 @@
...
+command = f'provenance.py -b {opensearch_endpoint} -l provenance.log -L DEBUG'
+if username is not None:
+    command += f' -u {username} -p {passwd}'
+
+result = 0
+if remotesLists is None:
+    result = os.system(command)
+else:
+    for remoteList in remotesLists:
+        result = os.system(command + f' -c {remoteList}')
```
tloubrieu-jpl, line 84: Since everything is coded in python would it be easier to call provenance.py though a python function rather than in command line ?

support/provenance/terraform/variables.tf
```diff
@@ -0,0 +1,65 @@
...
+}
+
+variable "aws_fg_image" {
+  description = "AWS image name for Fargate"
+}
+
+variable "aws_ecr_pull_policy_arn" {
+  description = "ARN for the ECR pull policy"
+  default = "arn:aws:iam::aws:policy/service-role/AWSAppRunnerServicePolicyForECRAccess"
+}
```
tloubrieu-jpl, line 65: Thanks for providing the terraform scripts here, that looks good. Did the SA agreed to have them on our public github ?

tloubrieu-jpl (approved):
Some minor comments or questions. Thanks Jimmie

Issue #252, opened 2023-02-21
Invalid LidvidsContext bounds fails internally and returns http500, vs expected handle and http400


alexdunnjpl:
## 💡 Description
ex.
http://localhost:8080/collections/urn:nasa:pds:mars2020_supercam:data_calibrated_audio::4.0/products?start=0&limit=-1
http://localhost:8080/collections/urn:nasa:pds:mars2020_supercam:data_calibrated_audio::4.0/products?start=-1&limit=1
http://localhost:8080/collections/urn:nasa:pds:mars2020_supercam:data_calibrated_audio::4.0/products?start=-1&limit=-1

All three should throw during `URIParameters` instantiation, with exceptions being caught and handled.

alexdunnjpl:
@al-niessner @nutjob4life @tloubrieu-jpl @jimmie are the [javax API parameter validation annotations](https://github.com/NASA-PDS/registry-api/blob/main/service/src/main/java/gov/nasa/pds/api/registry/controller/SwaggerJavaClassesTransmuter.java#L93-L94) purely for documentation purposes, or are they expected to trigger enforcement behaviour?

I'm not sure whether the validation isn't working, or whether I need to roll my own handler for IllegalArgumentExceptions somewhere up the chain.

alexdunnjpl:
On hold pending additional info

al-niessner:
@alexdunnjpl @tloubrieu-jpl 

It has been my experience that they URL parameter limits are enforced - limit as a letter did not work. Also limit used to be 1 to infinity and I changed it to 0 to infinity to get it working.

That said, I know the code generator is being changed so I do not know for certain anymore.

alexdunnjpl:
Thanks Al.

It doesn't appear that they're being respected/enforced, so that's something I'll need to follow up on I think.

Unrelated relevant note:

>  * 3. The default values for limit and start work in conjunction later in the
>  *    business logic to indicate a Singular or Plural return type. See swagger.yml
>  *    to understand Singular and Plural return types.

This indicates that stripping out `isSingular()` as a side-effect of excising `isSummaryOnly()` et al. per @tloubrieu-jpl 's input (Slack) may break something, and it may be necessary to understand and reimplement that functionality. 

It looks as though the unintuitive `URIParameters.start` default value of `-1` is used as some kind of flag to indicate whether the API endpoint is a Singular or Plural (i.e. collection?) return.

alexdunnjpl:
[...]

PR #256, opened 2023-02-23
Miscellaneous internal changes

### LINKED ISSUE #255: Incorporate misc changes loosely-related to #252
## 💡 Description
Preliminary work undertaken as part of #252 which does not actually resolve it (since it has been deprioritized but the work is still useful).
 - Add internal `IllegalArgumentException` to invalid `start` and `limit` setter calls
 - Remove summaryOnly-related internal code and rework Singular/Plural API route behaviour to be more explicit

alexdunnjpl:
## 🗒️ Summary
Preliminary work undertaken as part of #252 which does not actually resolve it (since it has been deprioritized but the work is still useful).
 - Add internal `IllegalArgumentException` to invalid `start` and `limit` setter calls
 - Remove summaryOnly-related internal code and rework Singular/Plural API route behaviour to be more explicit

## ⚙️ Test Data and/or Report
See Github Actions tests

## ♻️ Related Issues
Fixes #255 
Related to #252

### DIFF
--- a/model/swagger.yml
+++ b/model/swagger.yml
@@ -346,7 +346,7 @@ paths:
         - products
       summary: |
         returns all product versions with the identifier (lid) selected from all PDS Products. See identifier for details on how it determines what is returned.
-        
+
         note: if given a lidvid, it will be translated to a lid
       operationId: select-by-lidvid-all
       responses:
@@ -1107,7 +1107,7 @@ components:
       in: path
       description: |
         syntax: one of the allowable enum values
-        
+
         This is a shortcut for doing a query with ...
       required: true
       schema:
@@ -1149,8 +1149,6 @@ components:
         syntax: limit=10
 
         behavior: maximum number of matching results returned, for pagination
-        
-        note: limit=0 returns just the summary
       required: false
       schema:
         type: integer
@@ -1197,7 +1195,7 @@ components:
       in: path
       description: |
         syntax: one of the allowable enum values
-        
+
         This is a shortcut for doing a query with ...
       required: true
       schema:
@@ -1408,7 +1406,7 @@ components:
           xml:
             prefix: 'pds_api'
             namespace: 'http://pds.nasa.gov/api'
-        observing_system_components: 
+        observing_system_components:
           type: array
           description: list of instruments or platforms generating the data
           items:
@@ -1416,7 +1414,7 @@ components:
           xml:
             prefix: 'pds_api'
             namespace: 'http://pds.nasa.gov/api'
-        targets: 
+        targets:
           type: array
           description: list of targets or feature of interest the observation.
           items:
@@ -1475,7 +1473,7 @@ components:
           items:
             $ref: '#/components/schemas/wyriwygProduct'
     propertyArrayValues:
-      type: array 
+      type: array
       items:
         type: string
       xml:
--- a/service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java
@@ -27,7 +27,7 @@ import gov.nasa.pds.api.registry.search.RequestBuildContextFactory;
  *    of the set calls collocated.
  * 3. By default, the context assumes a Singular (single-element-valued) return type for the API route.
  *    If the route is a Plural (collection-valued) return type, the transmuter will call either/both of setStart() and
- *    setLimit(), which will mutate returnSingularDatum to its correct false value.
+ *    setLimit(), which will mutate singletonResultExpected to its correct false value.
  *    See swagger.yml for further detail of the Singular and Plural return types.
  */
 class URIParameters implements UserContext
@@ -41,7 +41,7 @@ class URIParameters implements UserContext
 	private PdsProductIdentifier productIdentifier = null;
 	private Integer start = 0;
 	private Integer limit = 0;  // Actual default value is passed in from the upstream frames of the call stack, but it's unclear where it comes from. Not swagger.yml, at least.
-	private Boolean returnSingularDatum = true;
+	private Boolean singletonResultExpected = true;
 	private String query = "";
 	private ProductVersionSelector selector = ProductVersionSelector.LATEST;
 	private List<String> sort = new ArrayList<String>();
@@ -60,7 +60,7 @@ class URIParameters implements UserContext
 	@Override
 	public Integer getLimit() { return limit; }
 	@Override
-	public boolean getReturnSingularDatum() { return returnSingularDatum; }
+	public boolean getSingletonResultExpected() { return singletonResultExpected; }
 	@Override
 	public String getLidVid() { return productIdentifier != null ? productIdentifier.toString() : ""; }
 	@Override
@@ -110,7 +110,7 @@ class URIParameters implements UserContext
 		}
 
 		this.limit = limit;
-		this.returnSingularDatum = false;
+		this.singletonResultExpected = false;
 		return this;
 	}
 	public URIParameters setProductIdentifier(ControlContext control) throws IOException, LidVidNotFoundException
@@ -139,7 +139,7 @@ class URIParameters implements UserContext
 		}
 
 		this.start = start;
-		this.returnSingularDatum = false;
+		this.singletonResultExpected = false;
 		return this;
 	}
 	public URIParameters setVerifyClassAndId (boolean verify)
--- a/service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java
@@ -25,14 +25,13 @@ import gov.nasa.pds.api.registry.search.RequestBuildContextFactory;
  *    a bunch of set calls. Instead of a single line for each set, then can be
  *    concatenated via the . to make the code more readable by keeping the all
  *    of the set calls collocated.
- * 3. The default values for limit and start work in conjunction later in the
- *    business logic to indicate a Singular or Plural return type. See swagger.yml
- *    to understand Singular and Plural return types.
+ * 3. By default, the context assumes a Singular (single-element-valued) return type for the API route.
+ *    If the route is a Plural (collection-valued) return type, the transmuter will call either/both of setStart() and
+ *    setLimit(), which will mutate returnSingularDatum to its correct false value.
+ *    See swagger.yml for further detail of the Singular and Plural return types.
  */
 class URIParameters implements UserContext
 {
-	private static final int SUMMARY_SAMPLE_SIZE = 100;
-
 	private boolean verifyClassAndId = false;
 	private String accept = "application/json";
 	private List<String> fields = new ArrayList<String>();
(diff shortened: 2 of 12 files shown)


model/swagger.yml
```diff
@@ -1149,8 +1149,6 @@ components:
         syntax: limit=10
 
         behavior: maximum number of matching results returned, for pagination
-        
-        note: limit=0 returns just the summary
```
tloubrieu-jpl, line 1153: I'll keep that note, although data is still returned with length 0, I would like that we made obvious to the users that limit=0 is the way to get the summary only.

service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java
```diff
@@ -103,21 +102,15 @@ public URIParameters setKeywords(List<String> keywords)
...
-			}
+		if (limit == null) {return this;}
+
+		if (limit < 0) {
+			String errMsg = String.format("start index must be 0 or higher (got '%d'))", start);
+			throw new IllegalArgumentException(errMsg);
 		}
+
+		this.limit = limit;
+		this.returnSingularDatum = false;
```
tloubrieu-jpl: Somehow the property name is not meaningful to me. A bit too abstract maybe.

service/src/main/java/gov/nasa/pds/api/registry/model/Pds4ProductBusinessObject.java
```diff
@@ -93,11 +93,8 @@ public int setResponse(HitIterator hits, Summary summary, List<String> fields, b
         {
             uniqueProperties.addAll(ProductBusinessObject.getFilteredProperties(kvp, fields, null).keySet());
 
-            if (!onlySummary)
-            {
-            	Pds4Product prod = Pds4ProductFactory.createProduct(hits.getCurrentId(), kvp, this.isJSON);
-            	list.add(prod);
-            }
+			Pds4Product prod = Pds4ProductFactory.createProduct(hits.getCurrentId(), kvp, this.isJSON);
```
tloubrieu-jpl, line 94: why is the identation so far on the right ?

service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
```diff
@@ -143,8 +143,9 @@ private RequestAndResponseContext(
 	public final GroupConstraint getPresetCriteria() { return this.presetCriteria; };
 	public ProductVersionSelector getSelector() { return this.selector; }
 
-	public boolean isSingular() { return this.getStart() == -1 && this.isSummaryOnly(); } // isSummaryOnly implies original limit value == 0
-	public boolean isSummaryOnly() { return summaryOnly; }
+
+	public boolean isSingular() { return this.returnSingularDatum; }
```
tloubrieu-jpl: If I am not mistaken, I mean if I understand what you do with returnSingularDatum: we don't want to return a specific format when limit=0 or 1. The format singular or plural is defined by the end-point.

It is very likely that I don't understand what you do here, so we can discuss that later.

tloubrieu-jpl (commented):
Overall I like the simplification of the code, but I would like to make sure the behavior is still what we expect/want.

Thanks Alex

model/swagger.yml
alexdunnjpl: Is that useful information to provide to the user?  "If you request zero data you'll get zero data" seems pointless, and specifying that you'll get just the summary (which is actually the search metadata) seems to indicate that something nonstandard occurs.  Your call, though.

service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java
alexdunnjpl: This replaces `isSingular` (dragged down from `RequestAndResponseContext`).  The meaning is "this `URIParameters` is being used for an API route which returns a single element (ex. `/bundles/{identifier}`) rather than a collection of elements (ex. `/bundles/{identifier}/products`)".

Open to suggestions on this as you're right - it's not great.

service/src/main/java/gov/nasa/pds/api/registry/model/Pds4ProductBusinessObject.java
alexdunnjpl: Probably something I missed, and I haven't got my autoformatter set up yet - will fix

service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
alexdunnjpl: Previously, there was some slightly-murky (to someone who isn't already familiar with it) state checking of (values pulled from) the `URIParameters` going on to determine whether the API route was a return-element or return-collection type.

That implicit logic has been made explicit and encapsulated within the `URIParameters` object.

The logic itself has not changed meaningfully - if the setters for `start` or `limit` are called on `URIParameters` object, the flag changes from return-element (returnSingularDatum=True) to return-collection(returnSingularDatum=False).  I didn't want to overcomplicate the changes, but if this is sufficiently unclear, a potential improvement would be to change the relevant member from `boolean isSingle` to an enumerated type `routeReturnType` with values `SingleElement` and `ElementCollection` or something similar.

service/src/main/java/gov/nasa/pds/api/registry/model/Pds4ProductBusinessObject.java
alexdunnjpl: Actually this wasn't a new change - there's some mixing of spaces and tabs throughout this file.  I've reformatted that function, but can expand the scope if desired.

model/swagger.yml
jordanpadams: @tloubrieu-jpl ☝️

service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java
jordanpadams: @tloubrieu-jpl @alexdunnjpl ☝️

nutjob4life (approved):

service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java
tloubrieu-jpl: @alexdunnjpl , What about something like `expectSingularResponse` ?

model/swagger.yml
tloubrieu-jpl: @alexdunnjpl we wanted a way for the user to get the metadata for a set of results, beyond the pagination limits, without requesting them all. That is useful for a user who would like to know the set of properties which can be requested in this products.

The use case is:
1. user want to explore the product's of a collection (e.g. 100000s products)
2. the user needs to know which properties are used to describe these products
3. the user can apply search criteria on some of the available properties to subset the collection of products (these properties cannot be guessed by the user)

That is for step 2 that we initially defined the `summaryOnly` keyword, and then we decided to remove this keyword to use limit=0 instead. That being said it sounds like the summary does not provide the description for the full request results beyond pagination limits but only for the current page.

alexdunnjpl: @tloubrieu-jpl if I understand you correctly, it sounds like `limit=0` _is_ supposed to behave in a different/non-standard manner, then? Specifically, to provide the union of the set of properties present in each product for the given query? (That's a non-homogeneous collection of property sets, yeah? Otherwise the user could just get it with `limit=1` and extract the property set themselves)

It currently doesn't, iirc, so sounds like there's a need to make some kind of fix if that's necessary, but that almost seems like a job for a subroute like `products/properties?{someQuery}` rather than serving it implicitly in the case of `products?{someQuery}&limit=0`?  This also seems like a potentially-expensive operation in the case of large query results, not that there's any getting around that if it's needed.

service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java
alexdunnjpl: I think it's slightly better, but I'm starting to feel like if it's worth worrying about, it's worth doing properly ([see the last bit of the comment in the next thread](https://github.com/NASA-PDS/registry-api/pull/256#discussion_r1116109721)), so unless you have an objection I think that I should 
 - make it an enum type instead of a boolean
 - put an explanatory comment in the enum type so that anyone who wonders/cares what exactly it refers to can jump to the explanation with a minimum of effort

alexdunnjpl: Actually, how about `singletonResultExpected`?

Communicates that the expected singularness applies to the _result_ (of the query), and I think that "singleton" makes the connection a little clearer because it either means "a unique instance" or "a set with cardinality 1", both of which are correct in this context.

tloubrieu-jpl (approved):
I fixed the conflict (that I introduced through an update on main branch).
I ran the integration test and I don't see an impact, before merging that would be good to have a review from @al-niessner.

tloubrieu-jpl:
@al-niessner could you review this PR ? Thanks

alexdunnjpl:
@al-niessner @tloubrieu-jpl still a couple of outstanding convos (above) before finalization, FYI

service/src/main/resources/application.properties
```diff
@@ -1,6 +1,6 @@
 springfox.documentation.swagger.v2.path=/api-docs
 server.contextPath=/
-server.port=8080
```
alexdunnjpl: Need to remove changes to this file before merge (note to self)

alexdunnjpl: resolved in 6b37541

al-niessner (approved):

alexdunnjpl:
Squash when merging, with following commit message (we've lost the commit history, but at least we can retain the granular messages):

* add IllegalArgumentException throws to URIParameters setStart() and setLimit() methods
* excise summaryOnly-related behaviour
* rework Singular/Plural status of API routes
  this is more explicit than the previous approach
* remove illegal character
  presence of emoji breaks branch integration testing

service/src/main/java/gov/nasa/pds/api/registry/controller/URIParameters.java
alexdunnjpl: @tloubrieu-jpl tentatively resolved by 90c4fb91f1bbade1efdfc50058ab6a2945285070 (if you agree with the new naming)

service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
alexdunnjpl: @tloubrieu-jpl unless I'm missing something, this is resolved - feel free to re-open if you have outstanding questions

model/swagger.yml
tloubrieu-jpl: @alexdunnjpl I was unclear, I am not looking for a non standard management of limit=0. I believe what is rather needed to be updated is the way the summary is prepared. For the use case I described above, the summary need to be made from the products matching the query (q, keyword) but ignoring the pagination parameters (start, limit). I will make a more detailed ticket for that, because all the summary properties are not equivalent.

PR #259, opened 2023-02-24
Reintroduce lidvid not found case when testing that the latest detection script runs

### LINKED ISSUE #258: not found lidvid does not return 404 error
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

On the reference test registry (see registry repository)
When I do request:

curl --request GET 'http://localhost:8081//products/urn:nasa:pds:insight_rad::2.1' \
--header 'Accept: application/vnd.nasa.pds.pds4+json'

I get a 500 error instead of 404

### 🕵️ Expected behavior

I expected to get a 404 error, lidvid not found

### 📜 To Reproduce

Run the integration tests

### 🖥 Environment Info

_No response_

1.2.0-SNAPSHOT, branch main

_No response_

### 🦄 Related requirements

🦄 #xyz

### ⚙️ Engineering Details

_No response_

tloubrieu-jpl:
## 🗒️ Summary
Very simple change

## ⚙️ Test Data and/or Report
The test is part of the standard integration tests found in repository 'registry'

## ♻️ Related Issues

Fixes #258

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
@@ -357,7 +357,7 @@ public class RequestAndResponseContext implements RequestBuildContext, RequestCo
         long hitCount = hits.getTotalHits().value;
         if (hitCount == 1L) {
           this.formatters.get(this.format).setResponse(hits.getAt(0), this.fields);
-        } else {
+        } else if (hitCount > 1L) {
           String basicErrMsg =
               "Got " + hitCount + " hits for a query which should have returned a singular result. "
                   + "Is provenance metadata present and up-to-date?";
--- a/service/src/main/resources/application.properties
+++ b/service/src/main/resources/application.properties
@@ -1,6 +1,6 @@
 springfox.documentation.swagger.v2.path=/api-docs
 server.contextPath=/
-server.port=8080
+server.port=8081
 server.forward-headers-strategy=framework
 
 #spring.jackson.date-format=io.swagger.RFC3339DateFormat
@@ -18,7 +18,7 @@ server.ssl.key-store=classpath:keystore.p12
 server.ssl.key-store-type=PKCS12
 
 # note the port is mandatory even when it is default :80 or :443
-openSearch.host=localhost:9200
+openSearch.host=elasticsearch:9200
 openSearch.registryIndex=registry
 openSearch.registryRefIndex=registry-refs
 openSearch.timeOutSeconds=60
--- a/service/src/main/resources/application.properties
+++ b/service/src/main/resources/application.properties
@@ -1,6 +1,6 @@
 springfox.documentation.swagger.v2.path=/api-docs
 server.contextPath=/
-server.port=8081
+server.port=8080
 server.forward-headers-strategy=framework
 
 #spring.jackson.date-format=io.swagger.RFC3339DateFormat
@@ -18,7 +18,7 @@ server.ssl.key-store=classpath:keystore.p12
 server.ssl.key-store-type=PKCS12
 
 # note the port is mandatory even when it is default :80 or :443
-openSearch.host=elasticsearch:9200
+openSearch.host=localhost:9200
 openSearch.registryIndex=registry
 openSearch.registryRefIndex=registry-refs
 openSearch.timeOutSeconds=60


service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
```diff
@@ -357,7 +357,7 @@ public void setResponse(RestHighLevelClient client, SearchRequest request) throw
         long hitCount = hits.getTotalHits().value;
         if (hitCount == 1L) {
           this.formatters.get(this.format).setResponse(hits.getAt(0), this.fields);
-        } else {
+        } else if (hitCount > 1L) {
```
alexdunnjpl, line 360: Unless I'm missing something, the logic change here seems incorrect - the idea of that set of conditionals is "if the response yields a single result, as expected, throw an error, otherwise there is an incorrect number of responses (either zero, or multiple), in which case a descriptive error should be thrown".

If this change addresses a test failure, the test case should be adjusted instead imho.

service/src/main/resources/application.properties
```diff
@@ -1,6 +1,6 @@
 springfox.documentation.swagger.v2.path=/api-docs
```
alexdunnjpl: Seems like this file was attached to the commit erroneously - suggest removal of this change.

alexdunnjpl (changes requested):

service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
tloubrieu-jpl: 0 is not an error since it can happen that the lidvid is not found in the database, in which case we want to return 404 error and not 500 with the message on latest script. The change I made makes the not found lidvid (0 found) return 404 again (tested with integration tests)

I can not be positive though that my approach is the best regarding the overall architecture of the application, @al-niessner would help us on that.

service/src/main/resources/application.properties
tloubrieu-jpl: I removed these local updates fr my configuration, thanks.

service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
alexdunnjpl: @tloubrieu-jpl what you're saying makes sense, but let me take a closer look to make sure that this isn't in a context where exactly-one hit is expected - I recall changing this conditional a couple of weeks back and it doesn't seem like a simple/understandable mistake for me to have messed up the logic so badly during that change.

alexdunnjpl: Here's the commit in question if you want to get in before I do

https://github.com/NASA-PDS/registry-api/commit/22a0970d2f23046ba07842c21da2076cebab16fc?diff=split

tloubrieu-jpl:
@al-niessner could you review this PR ?

service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
tloubrieu-jpl: @alexdunnjpl it does not look like your commit introduced that error, since before, the `else` clause was already both referring to cases with 0 or >1 results. The change I am proposing it to exclude the case 0 result from the `else`.

I am not questioning the behavior since after my PR, we are getting the expected behavior, but there might be a better way, especially more readable way of doing that.

alexdunnjpl: @al-niessner could you chime in on whether the original logic (see 22a0970 pre-diff version) is a mistake, or correct given some context @tloubrieu-jpl and I are missing?

alexdunnjpl:
Branch test failure is due to bug fixed in https://github.com/NASA-PDS/registry-api/pull/256/commits/ce237d3a44fa9d9472193662f3f4abce9ab9313e which is already merged to `main`.  Suggest cherry-picking that commit to confirm that BITs pass prior to merge.

tloubrieu-jpl:
@alexdunnjpl , since @al-niessner is mostly unavailable this week, I suggest to merge this PR (since it does what I expect).

PR #263, opened 2023-02-27
261: lookup product class when group not explicitly given

### LINKED ISSUE #261: The members of a bundle can not be requested
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When I want to see the members of a bundle:

    curl --request GET 'http://localhost:8081//products/urn:nasa:pds:insight_rad::2.1/members' \
--header 'Accept: application/json'

or 

        curl --request GET 'http://localhost:8081//products/urn:nasa:pds:insight_rad::2.1/members/members' \
--header 'Accept: application/json'

I get an error:

{
    "request": "//products/urn:nasa:pds:insight_rad::2.1/members",
    "message": "The lidvid 'urn:nasa:pds:insight_rad::2.1' of type product does not support members membership"
}

### 🕵️ Expected behavior

I expected to get the collections (/members) or the observational products/documents (/memebers/members) of the bundle 

### 📜 To Reproduce

See integration test cases in repository `registry`

### 🖥 Environment Info

_No response_

1.2.0-SNAPSHOT on main branch

_No response_

### 🦄 Related requirements

🦄 #https://github.com/NASA-PDS/registry-api/issues/223

### ⚙️ Engineering Details

_No response_

### LINKED ISSUE #268: product's members does not work on a collection
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When I did curl request:

    curl 'http://localhost:8081//products/urn:nasa:pds:insight_rad:data_derived::7.0/members'

I am getting the response:

```json
    {
    "request": "//products/urn:nasa:pds:insight_rad:data_derived::7.0/members",
    "message": "The lidvid 'urn:nasa:pds:insight_rad:data_derived::7.0' of type product does not support members membership"
}
```

### 🕵️ Expected behavior

I expected to get the list of observational products or documents belonging to this collection

### 📜 To Reproduce

Use the integration test setup proposed in the registry repository.

### 🖥 Environment Info

_No response_

1.2.0-SNAPSHOT on main branch

Note that same error occurs with the member-of suffix

### 🦄 Related requirements

🦄 #https://github.com/NASA-PDS/pds-api/issues/58

### ⚙️ Engineering Details

_No response_

al-niessner:
## 🗒️ Summary
When using products as general, also need to parse the product class from the DB because the expected group is no longer specified (this is a good thing). Modified the member transmuter to handle this condition.

## ♻️ Related Issues
Closes #261 
Closes #268

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,7 +35,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 
 <!--
-👉 This is the TEMPLATE project object model or `pom.xml` file.
+This is the TEMPLATE project object model or `pom.xml` file.
 
 Go through this file line-by-line and replace the template values with your own.
 -->
@@ -225,7 +225,7 @@ Go through this file line-by-line and replace the template values with your own.
                                 <goals>
                                     <goal>jar</goal>
                                     <goal>test-jar</goal>
-                                </goals> 
+                                </goals>
                             </execution>
                         </executions>
                     </plugin>
--- a/service/src/main/java/gov/nasa/pds/api/registry/controller/Member.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/controller/Member.java
@@ -15,6 +15,7 @@ import gov.nasa.pds.api.registry.exceptions.NothingFoundException;
 import gov.nasa.pds.api.registry.exceptions.UnknownGroupNameException;
 import gov.nasa.pds.api.registry.model.ReferencingLogicTransmuter;
 import gov.nasa.pds.api.registry.model.RequestAndResponseContext;
+import gov.nasa.pds.api.registry.search.QuickSearch;
 
 class Member implements EndpointHandler
 {
@@ -30,7 +31,11 @@ class Member implements EndpointHandler
 			throws ApplicationTypeException, IOException, LidVidNotFoundException, MembershipException,
 			NothingFoundException, UnknownGroupNameException
 	{
-		ReferencingLogic transmuter = ReferencingLogicTransmuter.getBySwaggerGroup(content.getGroup()).impl();
+		ReferencingLogic transmuter;
+		
+		if (0 < content.getGroup().length()) transmuter = ReferencingLogicTransmuter.getBySwaggerGroup(content.getGroup()).impl();
+		else transmuter = ReferencingLogicTransmuter.getByProductClass(QuickSearch.getValue(control.getConnection(), false, content.getLidVid(), "product_class")).impl();
+
 		RequestAndResponseContext context = this.offspring ? transmuter.member(control, content, this.twoSteps) :
 			                                                 transmuter.memberOf(control, content, this.twoSteps);
 		return new ResponseEntity<Object>(context.getResponse(), HttpStatus.OK);
--- a/service/src/main/java/gov/nasa/pds/api/registry/model/RefLogicBundle.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/model/RefLogicBundle.java
@@ -10,6 +10,7 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.stream.Collectors;
 
+import gov.nasa.pds.api.registry.RequestBuildContext;
 import gov.nasa.pds.api.registry.model.identifiers.LidVidUtils;
 import gov.nasa.pds.api.registry.model.identifiers.PdsLidVid;
 import gov.nasa.pds.api.registry.model.identifiers.PdsProductIdentifier;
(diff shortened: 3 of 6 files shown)


al-niessner:
@tloubrieu-jpl 

Should be working now.

al-niessner:
@jordanpadams @nutjob4life @tloubrieu-jpl 

Ready for review

nutjob4life (approved):
👍

tloubrieu-jpl (changes requested):
Hi Al,

Now the request:

    curl 'http://localhost:8081/products/urn:nasa:pds:insight_rad::2.1/members' --header 'Accept: application/json'

Returns an empty result: 
`    {"summary":{"q":"","hits":0,"took":13,"start":0,"limit":100,"sort":[],"properties":[]},"data":[]}`

But we should have a few collections returned with the reference test dataset.

alexdunnjpl:
@tloubrieu-jpl @al-niessner 

[If a LIDVID referenced by the aggregate product is missing from the registry, this will result in a LidVidNotFoundException.](https://github.com/NASA-PDS/registry-api/blob/39ffefe9cbd7abae32d9259023f6a79a438ab100/service/src/main/java/gov/nasa/pds/api/registry/model/identifiers/LidVidUtils.java#L54) and prevent some/all of the found LIDVIDs from being included in the results returned to the user.

[Such exceptions are caught and logged](https://github.com/NASA-PDS/registry-api/blob/39ffefe9cbd7abae32d9259023f6a79a438ab100/service/src/main/java/gov/nasa/pds/api/registry/model/RefLogicBundle.java#L171) and do not bubble up to where exceptions are turned into non-200 HTTP responses  (this behaviour also occurs in `RefLogicCollection` and `RefLogicProduct` - originally in LidVidUtils per 341e31b0)

A decision needs to be made on whether it's still (ever) desirable to catch and log such errors without re-throwing.  If not, re-throws should be added to each of the three catch blocks, or 341e31b0 should be reverted and a re-throw added to the original location in `LidVidUtils`.  That commit was only ever made to allow future callers to access any resulting `LidVidNotFoundExceptions` rather than the utility function swallowing them.

I've confirmed that the absence of collection `urn:nasa:pds:insight_rad:data_calibrated` from the test dataset is the superficial cause of the erroneous behaviour.

tloubrieu-jpl:
@alexdunnjpl I am missing the context for the commit https://github.com/NASA-PDS/registry-api/commit/341e31b0b0c40e93aa4b21a6d6328fa2338ef779 but I would tell you that:
if a member of a collection or bundles referenced in its relationships, an exception should be thrown and appear in the log (WARNING) but the remaining members should be returned to the user. At least that is how I am thinking now.

Otherwise we should have a more explicit error message, saying some of the `members` are missing in the registry and the service was not able to retrieve the full response. I am not sure how it is best to provide a response to the user with warning, but let's not go this route now.

alexdunnjpl:
@tloubrieu-jpl the issue is that if I stop swallowing the error, it might end up presenting in other contexts where it wasn't desirable.  Small chance, though, and if it does we can resolve that at that stage.

So you're happy for me to add the re-throw of that error, then?

alexdunnjpl:
@tloubrieu-jpl isn't it sketchy to serve up known-incomplete data to the user?  Currently, if the exception is allowed to bubble up, they're presented with a "$LIDVID does not exist in the registry", which at least tells them what they need to do to successfully make the call (harvest the missing product in question).

I can implement it as-described if you're sure, though.

tloubrieu-jpl:
I confirm that with the branch issue_261, I am getting this error message in logs:

    2023-03-07 16:34:20.212 ERROR 43661 --- [nio-8081-exec-1] g.n.p.api.registry.model.RefLogicBundle  : Database referential integrity error -  LID is referenced but does not exist in db: 

gov.nasa.pds.api.registry.exceptions.LidVidNotFoundException: The lidvid Could not find any LIDVIDs for LID urn:nasa:pds:insight_rad:data_calibrated was not found

and the response gives an empty result, which might be what I asked for in my previous comment. I will confirm that by looking into the test database. I believe we need to understand better why these integrity errors appears and then what we should do about it.

alexdunnjpl:
@tloubrieu-jpl implemented as-requested (log warning, return retrievable elements of intended results set)

Manually tested `members`, `members/members`, `member-of` and `member-of/member-of`.

tloubrieu-jpl:
So @alexdunnjpl , in this case the collection is available in the registry (see request on your local test deployment https://elasticsearch:9200/registry/_doc/urn:nasa:pds:insight_rad:data_calibrated::7.0) but its status is `staged` which makes it invisible to the API users.

I believe we should definitely then return all the other responses. Another case I am thinking of is a collection where products are added incrementally over time. The latest products arrived might be in a stage status, but the user would need to access all the other products collected so far and in `archived` status.

So for ticket 261 I am thinking of 2 fixes:
- in this case change the error message to warning in logs, saying `one LID referenced in the product {LID} is not available in the registry, possibly because of its visibility status` instead of `Database referential integrity error -  LID is referenced but does not exist in db`
- raise the warning so that it does not prevent the API from returning the rest of the members.

For later, we can create a ticket where we will inform the user that some of the members where not retrieved. See https://github.com/NASA-PDS/registry-api/issues/270

alexdunnjpl:
@tloubrieu-jpl Done(ish - let me know if you still prefer your suggested log message to the updated one).

tloubrieu-jpl:
Thanks @alexdunnjpl I tested the request and it work as I expected. I still would like the log message to be update since it is not an integrity error.
The message should be:
`The LID {LID not found} referenced in the product {LID parent} is not found in the registry, possibly because of its archive status`

Thanks,

Thomas

tloubrieu-jpl:
Sorry I misread the error message provided by @alexdunnjpl's code update and what is proposed works for me.

tloubrieu-jpl (approved):

PR #264, opened 2023-02-28
Add logging messages to CF api rewrite function.

jimmie:
## 🗒️ Summary
Add logging messages to CF funcxtion which provide incoming URI and resulting rewrite (if applicable)

## ♻️ Related Issues
* https://github.com/NASA-PDS/cloud-tasks/issues/31

Log output will go to a cloudwatch log group /aws/cloudfront/function/api_uri_rewrite in the **us-east-1** region

### DIFF
--- a/service/terraform/aws/api_uri_rewrite.js
+++ b/service/terraform/aws/api_uri_rewrite.js
@@ -14,6 +14,7 @@ var VER_IDX = 3
 var CMD_IDX = 4
 
 var TARGET_URI_PREFIX = "/api/search"
+var DEBUG_PARAM       = "cf_debug" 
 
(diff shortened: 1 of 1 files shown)


alexdunnjpl (approved):
@jimmie might be worth making the semantics of the conditional on/near L37 explicit while you're here, but otherwise LGTM

`(uriParts.length > CMD_IDX && uriParts[CMD_IDX].trim() != "")`

jimmie:
This latest rev includes conditional logging based on the presence (value is disregarded) of 'cf_debug' as a query parameter. It is removed prior to submission to the load balancer if present.

PR #274, opened 2023-03-09
Fix mvc error introduced by management of the Accept:* header

tloubrieu-jpl:
## 🗒️ Summary
The XML formats were not rendered properly but as some kind of JSON since it became the default format.

I changed the order in which the message converters so that the default JSON is only used if none of the specific one defined before matches the user request (Accept header) before.


jordanpadams (commented):
merging this one before #275

jordanpadams (approved):

PR #276, opened 2023-03-09
Invoke provenance script w/ --reset flag

jimmie:
## 🗒️ Summary
Add --reset flag to provenance execution in order to force recalculation of all document histories.

This PR also incorporates terraform changes to incorporate creation of EventBridge scheduler components.

### DIFF
--- a/support/provenance/terraform/provider.tf
+++ b/support/provenance/terraform/provider.tf
@@ -1,5 +1,5 @@
 provider "aws" {
-  version = ">= 4.7"
+  version = "~> 4.7"
   region  = var.aws_region
   profile = var.aws_profile
 }
--- a/support/provenance/terraform/provider.tf
+++ b/support/provenance/terraform/provider.tf
@@ -1,5 +1,5 @@
 provider "aws" {
-  version = "~> 3.0"
+  version = ">= 4.7"
   region  = var.aws_region
   profile = var.aws_profile
 }
--- a/support/provenance/terraform/secrets.tf
+++ b/support/provenance/terraform/secrets.tf
@@ -15,7 +15,7 @@ resource "aws_ssm_parameter" "prov_endpoint_parameter" {
   tags = {
     Alfa = var.node_name_abbr
     Bravo = var.venue
-    Charlie = "provenance"
+    Charlie = "registry-provenance"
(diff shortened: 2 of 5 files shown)


support/provenance/terraform/provider.tf
```diff
@@ -1,5 +1,5 @@
 provider "aws" {
-  version = "~> 3.0"
+  version = ">= 4.7"
```
jordanpadams: @jimmie per discussions we have been having with @nutjob4life about issues we have encountered with python version collisions, do we want to pin to a more specific version?

jordanpadams: @tloubrieu-jpl ☝️

nutjob4life: That "pessimistic constraint operator" `~>` is good stuff!

nutjob4life (commented):
I'm not at all familiar with Terraform (aside from the handy-dandy pessimistic constraint operator) so I can't approve/request changes—just comment 😇

support/provenance/terraform/secrets.tf
```diff
@@ -15,7 +15,7 @@ resource "aws_ssm_parameter" "prov_endpoint_parameter" {
   tags = {
     Alfa = var.node_name_abbr
     Bravo = var.venue
-    Charlie = "registry"
+    Charlie = "provenance"
```
tloubrieu-jpl: @jimmie we could prefix the application name by `registry` ? since this is definitelly a subcompoenent of it. However I don't know how these tags are used so I'll let you decide which is best. Thanks

support/provenance/terraform/provider.tf
tloubrieu-jpl: I've just learned how that works and I agree with what Sean said

tloubrieu-jpl (approved):
Except for my note on Charlie tag which could include `registry` as a prefix to make clear this components belong to the registry.

support/provenance/terraform/secrets.tf
jordanpadams: @jimmie ☝️

tloubrieu-jpl (approved):

PR #278, opened 2023-03-13
revert erroneous change to default server port

alexdunnjpl:
## 🗒️ Summary
Reverts change to default server port.  Presumably this was not intentional as the documentation still specifies 8080

No related issue

### DIFF
--- a/service/src/main/resources/application.properties
+++ b/service/src/main/resources/application.properties
@@ -1,5 +1,5 @@
 server.contextPath=/
-server.port=8081
+server.port=8080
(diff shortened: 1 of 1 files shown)


nutjob4life (approved):
Hmmm, I don't know. This change looks a bit dodgy 😅

alexdunnjpl:
@nutjob4life yeah, I really should've made the change more clear, but with such a complex and wide-ranging implementation I had to just do my best

tloubrieu-jpl (approved):

PR #289, opened 2023-03-15
revert erroneous change to default server config

alexdunnjpl:
Reverts the other unintended(?) change (see #278 ) to default configuration


tloubrieu-jpl (approved):
Sorry about that again, I need to find a way to keep a local configuration which is not commited on the repository but overrides the default configuration...

PR #294, opened 2023-03-22
fix Pds4ProductFactory.createDataFiles()

### LINKED ISSUE #293: Requests with Accept:application/vnd.nasa.gds.pds4+json fail for products with no ops:Label_File_Info/ops:file_name
### Checked for duplicates

No - I haven't checked

### 🐛 Describe the bug

When I performed `curl --get 'http://localhost:8080/products/urn:nasa:pds:orex.ovirs' --header 'Accept:application/vnd.nasa.pds.pds4+json'` (not a product in the I&T set, but this probably presents for any bundle/collection), API returned HTTP500.

### 🕵️ Expected behavior

I expect the call to successfully return the label data as JSON.

### 📜 To Reproduce

No need - on it.

### 🖥 Environment Info

_No response_

_No response_

_No response_

### 🦄 Related requirements

_No response_

### ⚙️ Engineering Details

_No response_

alexdunnjpl:
## 🗒️ Summary
No longer throws when FLD_DATA_FILE_NAME is not present in fieldMap and `Accept:application/vnd.nasa.gds.pds4+json` specified.

Gut feeling is that this affects all bundles/collections.
## ⚙️ Test Data and/or Report
Still need to replicate add Postman regression test for this use-case.

@tloubrieu-jpl @jordanpadams could you please confirm where tests should be added? ./service/src/test/resource/postman_collection.json, and an equivalent in the registry docker repo?  It's not totally clear from the docs.

## ♻️ Related Issues
fixes #293

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/model/Pds4ProductFactory.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/model/Pds4ProductFactory.java
@@ -13,7 +13,7 @@ import gov.nasa.pds.model.Pds4Product;
 
 /**
  * Creates Pds4Product object from opensearch key-value field map.
- * 
+ *
  * @author karpenko
  */
 public class Pds4ProductFactory {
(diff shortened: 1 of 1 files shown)


nutjob4life (approved):
👍

PR #302, opened 2023-03-29
registry-155 - Registry multi-tenancy

jimmie:
## 🗒️ Summary
Enable registry multi-tenancy : co-located documents from various nodes within a single Opensearch domain. Provide support for a 'node=<id>' query parameter to all API endpoints which filters documents having 'ops:Harvest_Info/ops:node_name' equal to that value.

Also included is the modified CloudFront function which now passes an HTTP header 'x-request-version' that identifies the API version. This should be consumed by the EC2 Load Balancer rules to direct to the appropriate service (currently ECS/Fargate) endpoint/listener.

THIS PR SHOULD NOT BE MERGED AT THIS POINT IN TIME but is provided to ensure that all bases wrt the service code have been considered.

refs https://github.com/NASA-PDS/registry/issues/155
refs https://github.com/NASA-PDS/registry-api/issues/304

### DIFF
--- a/model/swagger.yml
+++ b/model/swagger.yml
@@ -80,6 +80,7 @@ paths:
         - $ref: "#/components/parameters/Query"
         - $ref: "#/components/parameters/Sort"
         - $ref: "#/components/parameters/Start"
+        - $ref: "#/components/parameters/Node"
   /classes/{class}/{identifier}/members:
     get:
       tags:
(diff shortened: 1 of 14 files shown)


jordanpadams:
converted PR to Draft since it should not be merged yet

jordanpadams:
Closing this PR as we move to a different architectural approach

PR #313, opened 2023-04-11
de-duplicate hits with common id when a singular result is expected

### LINKED ISSUE #306: If >1 products with the same LIDVID are registered by different nodes, we should respond with one of them
## 💡 Description

https://pds.nasa.gov/api/search/1/products/urn:nasa:pds:orex.spice::12.0 returns the error `Too many or too few lidvids matched the request when it should have just been 1.`.

If OpenSearch actually works with this collision, we should randomly pick one of them and this error should be alerted to the EN team, NOT to the user.

Note: Interim solution before implementing a more ubiquitous solution per https://github.com/NASA-PDS/registry/issues/93

alexdunnjpl:
this prevents products present in multiple clusters from causing a breaking error

## 🗒️ Summary
per title

## ⚙️ Test Data and/or Report
Manually tested.  Full testing requires a multi-cluster setup to test against, and the change is pretty trivial.  @jordanpadams @tloubrieu-jpl is further testing required?

## ♻️ Related Issues
fixes #306

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/model/RequestAndResponseContext.java
@@ -9,6 +9,7 @@ import java.util.Map;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.client.RequestOptions;
 import org.opensearch.client.RestHighLevelClient;
+import org.opensearch.search.SearchHit;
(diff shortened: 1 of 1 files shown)


nutjob4life (approved):
LGTM 👍

alexdunnjpl:
Approved via Thomas' questionably-warranted faith in my attention to detail.

PR #322, opened 2023-04-21
implement /product-properties endpoint

### LINKED ISSUE #277: Product summary object has an incomplete "properties" set
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

Currently, per @tloubrieu-jpl, the properties enumeration contained in the response returned from an endpoint like `/products/?q=someQuery` only contains an enumeration for the products contained in the page, which may be incomplete.

Initially-proposed requirement: provide full enumeration of properties with every page.

Problem:  This is O(n^2) with the number of products, as each page of API results would then require the API to iterate over all (pages of/products in) hits returned by the db.

Proposed solution: Do not provide `properties` information in the endpoint used to access products.  Instead, implement an endpoint to explicitly request this information, like `/products/properties/?someQuery`, which reduces "get the full set of properties and get all the products" to O(n) and avoids the need to mess around with caching. 

### 🕵️ Expected behavior

I expect (tentatively, pending approval) 

 - [ ] `properties` object is not returned in summary object
 - [ ] `/products/properties` endpoint implemented
 - [ ] `/collections/properties` endpoint implemented
 - [ ] `/bundles/properties` endpoint implemented

alexdunnjpl:
## 🗒️ Summary
Implements a new endpoint, /properties, which returns metadata about the searchable fields which users may use in their raw queries (`q=`)

This is pulled directly from the `registry` OpenSearch index and is thus fast and reliable.

## ⚙️ Test Data and/or Report
~Need to add a Postman test for this, but that can wait until Monday~

Postman test added in registry PR

## ♻️ Related Issues
Fixes #277


alexdunnjpl:
@tloubrieu-jpl is the postman collection in service/src/test/resources actually supposed to contain any tests, or just a sample set of requests?  Can't tell if it's the latter or if I'm just missing something silly.

nutjob4life (approved):

model/swagger.yml
```diff
@@ -311,6 +311,26 @@ paths:
       - $ref: "#/components/parameters/Sort"
       - $ref: "#/components/parameters/Start"
 
+  /product-properties:
```
tloubrieu-jpl: I was thinking of `/properties` as end-points. /products-properties sounds redundant to me. However `/products/properties` would make sense but there might be difficulties for the implemention, since 'properties' would be considered as a lid or lidvid....

model/swagger.yml
```diff
@@ -311,6 +311,26 @@ paths:
...
+    get:
+      tags:
+        - 1. all products
+      summary: |
+        return a list of all possible searchable metadata fields for products published in the registry
+      operationId: product-properties-list
+      responses:
+        '200':
+          $ref: "#/components/responses/Plural"
```
tloubrieu-jpl: I think we should have a specific response format for the properties, Plural is for responses returning a list of products.

service/src/main/java/gov/nasa/pds/api/registry/controller/SwaggerJavaProductsTransmuter.java
```diff
@@ -112,4 +113,13 @@ public ResponseEntity<Object> selectByLidvidLatest(String identifier,
     return this.processs(new Standard(), this.uriParametersBuilder.setIdentifier(identifier)
         .setFields(fields).setVersion(ProductVersionSelector.LATEST).build());
   }
+
+  @Override
+  public ResponseEntity<Object> productPropertiesList() {
+
+    String registryIndexName = this.getConnection().getRegistryIndex();
```
tloubrieu-jpl: I don't know if that is here, but I was expecting the response to be a bit less OpenSearch/ElasticSearch specific, especially the types, keyword is not meaningful here. We need to find something more explicit outside the context of OpenSearch.

tloubrieu-jpl (changes requested):
I am not approving it yet because some point requires discussion.

model/swagger.yml
alexdunnjpl: Yeah, the overload with `/products/{identifier}` was the motivation for the (admittedly awkward) naming.  I think the only acceptable approach is to find the most semantically-clear root path name we can.  

`properties` is probably best - I was thinking it might get confusing if we later wanted to have equivalents for only bundles and collections, too, but those could be implemented under `/classes/{class}/properties` since those don't and won't ever have to deal with `/classes/{class}/{identifier}`

model/swagger.yml
alexdunnjpl: This was an oversight on my part, good catch.  Will fix.

service/src/main/java/gov/nasa/pds/api/registry/controller/SwaggerJavaProductsTransmuter.java
alexdunnjpl: What you're seeing is the result of me just dumping the document without further manipulation (because the fine details of what you/Jordan want from this endpoint are pending discussion).

We can map these however you like (for example, mapping instances of `keyword` to `string` since that's what a `keyword` is in a user search context).

model/swagger.yml
alexdunnjpl: meeting decision: `/properties` (unless working group suggests otherwise)

service/src/main/java/gov/nasa/pds/api/registry/controller/SwaggerJavaProductsTransmuter.java
alexdunnjpl: meeting decision (tentative): map OS types to `["string", "float", "integer", "boolean"]`

tloubrieu-jpl:
@alexdunnjpl regarding your question on test in postman, we are having a few. They can be found in the `Test` tab of each request. See screenshot:
image

model/swagger.yml
alexdunnjpl: addressed in 5120fad8

service/src/main/java/gov/nasa/pds/api/registry/controller/SwaggerJavaProductsTransmuter.java
alexdunnjpl: addressed in 5120fad8

model/swagger.yml
alexdunnjpl: addressed in 5120fad8

alexdunnjpl:
> @alexdunnjpl regarding your question on test in postman, we are having a few. They can be found in the `Test` tab of each request. See screenshot: image

@tloubrieu-jpl where are you pulling that from?  Loading [the service tests file from this repository](https://github.com/NASA-PDS/registry-api/blob/main/service/src/test/resources/postman_collection.json) doesn't populate the tests tab with anything.

tloubrieu-jpl:
Hi @alexdunnjpl , you are right the file you are using is obsolete and should be removed. The one to use is in the registry repository https://github.com/NASA-PDS/registry/blob/main/docker/postman/postman_collection.json
Sorry about that.

alexdunnjpl:
@tloubrieu-jpl please review at your leisure (I think we ironed out the outstanding questions?)

Type stuff is at the following locations:
https://github.com/NASA-PDS/registry-api/pull/322/files#diff-efe8b5acb4310cc66513c73b36c8564a1aac7efcfd581be3c8bf0c1411f190a1R1307
https://github.com/NASA-PDS/registry-api/pull/322/files#diff-7bb6eb16a239a0e1d6422eaf4cece3577003084597f76d4ec0c2b8bdc7e0a673R148

al-niessner (approved):

alexdunnjpl:
Bypassing requested-changes block due to more-recent approval by other code owner

PR #328, opened 2023-04-28
326 remove "products" from exposed classes

### LINKED ISSUE #326: Remove the 'products' in the /classes
## 💡 Description

`products` stands for anything except `bundles` and `collections`.

I don't remember of a use case for that. If it is confirmed I would prefer to remove it, because any use of the `products` keyword outside its PDS4 meaning (any digital artifact) will be confusing for the users.

alexdunnjpl:
## 🗒️ Summary
Per @tloubrieu-jpl in #326, "products" is confusing when overloaded to describe "products which are not aggregate products.

This PR removes 'products' from the results returned by endpoint `/classes`, and removes it as a enumerated class value for endpoint `/classes/{class}`.

It also performs significant renaming of related concepts within the API, specifically:

Relevant `Product`-prefixed classes classes are now named with `NonAggegateProduct` prefix instead.
GroupConstraint methods any(), all() and not() have been renamed to filter(), must() and mustNot().  These new names directly reflect the search concepts they conceptually map to, and should avoid confusion in future.

A method union() is implemented for `GroupConstraint` and later reverted.  I haven't written unit tests for it and can't be bothered since it ended up not being required, but it's left in the commit history just in case a use-case for it comes up in future.

## ⚙️ Test Data and/or Report
Postman tests pass.  Postman tests written and PR will be opened to add them to the registry repo's tests

## ♻️ Related Issues
fixes #326


alexdunnjpl:
Oh, and I deleted the obsolete/non-maintained postman tests file and replaced it with a readme so devs can find the real one.

al-niessner (approved):

PR #332, opened 2023-05-03
Update github actions for deprecation

alexdunnjpl:
## 🗒️ Summary
Addresses ::set-output deprecation
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

## ⚙️ Test Data and/or Report
Not tested - will require visual review, and testing after merge

## ♻️ Related Issues
https://github.com/NASA-PDS/devops/issues/43

### DIFF
--- a/.github/workflows/stable-cicd.yaml
+++ b/.github/workflows/stable-cicd.yaml
@@ -85,7 +85,7 @@ jobs:
                 name: 🫙 Jar and Tag Determination
                 id: jartag
                 run: |
-                    echo "jar_file=$(ls ./service/target/*.jar)" >> $GITHUB_OUTPUT
+                    echo "jar_file=$(find ./service/target/ -maxdepth 1 -regextype posix-extended -regex '.*/registry-api-service-[0-9]+\.[0-9]+\.[0-9]+(-SNAPSHOT)?\.jar')" >> $GITHUB_OUTPUT
                     echo "image_tag=$(echo ${{github.ref}} | awk -F/ '{print $NF}')" >> $GITHUB_OUTPUT
             -
                 name: 💳 Docker Hub Identification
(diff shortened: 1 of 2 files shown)


.github/workflows/stable-cicd.yaml
```diff
@@ -85,8 +85,8 @@ jobs:
                 name: 🫙 Jar and Tag Determination
                 id: jartag
                 run: |
-                    echo "::set-output name=jar_file::$(ls ./service/target/*.jar)"
-                    echo "::set-output name=image_tag::$(echo ${{github.ref}} | awk -F/ '{print $NF}')"
+                    echo "jar_file=$(ls ./service/target/*.jar)" >> $GITHUB_OUTPUT
```
nutjob4life: This should be `./service/target/registry-api-service-*.jar` because there are also `-bin.jar`, `-sources.jar`, `-docs.jar`, etc. You want it to be as specific as possible and match just one file—the one with the compiled code.

.github/workflows/unstable-cicd.yaml
```diff
@@ -80,7 +80,7 @@ jobs:
             -
                 name: 🫙 Jar File Determination
                 id: jarrer
-                run: echo "::set-output name=jar_file::$(ls ./service/target/*.jar)"
+                run: echo "jar_file=$(ls ./service/target/*.jar)" >> $GITHUB_OUTPUT
```
nutjob4life: Ditto

nutjob4life (changes requested):

.github/workflows/stable-cicd.yaml
alexdunnjpl: fixed

.github/workflows/unstable-cicd.yaml
alexdunnjpl: ditto

alexdunnjpl:
@nutjob4life changed as requested.  This has the side-effect of prepending the path (./service/target) to the value of the variable, but since it's being passed to a Docker `ADD` the extra content should be fine.

nutjob4life (approved):
😅👍

Issue #336, opened 2023-05-16
As a PDS operator, I want to know the health of the registry API service


jimmie:
This indication will be based on a new API endpoint specific to indicating the health of the service. The primary purpose of this is to provide a comprehensive assessment of if the service is healthy, to be used as the healthcheck for the service listener.

Sending a request to this endpoint will provide a summary of the following:

- Opensearch connectivity
- Springboot metrics, if possible 
- Non-sensitive config details: active remotes (in CCS context), active indices (in new MT context)
- JVM  information: Runtime.freeMemory(), Runtime.maxMemory(), Runtime.version()

Failures or certain levels of the above info will result in a non-200 return code.

The determined state of health will be:
- returned in a JSON response payload. This will be disregarded by the listener group healthcheck but will provide key information to operators if invoked directly
- recorded in the service log file (i.e. stdout/stderr which is recorded in CloudWatch logs)

Once available, we will need to update the terraform scripts to include this endpoint in the listener healthcheck definition.

### Acceptance Criteria

**Given** a nominal running Registry API and OpenSearch registry
**When I perform** a query of the `healthcheck/` endpoint
**Then I expect** to receive a 200 response and metadata indicating a successful running application

**Given** a running Registry API and OpenSearch registry, with an off-nominal state for the Registry API
**When I perform** a query of the `healthcheck/` endpoint
**Then I expect** to receive TBD response code(s) and applicable metadata

**Given** a running Registry API and Registry (OpenSearch), with an off-nominal state for the Registry (OpenSearch)
**When I perform** a query of the `healthcheck/` endpoint
**Then I expect** to receive TBD response code(s) and applicable metadata

### Sub-tasks
- https://github.com/NASA-PDS/registry-api/issues/350

jimmie:
Derived from registry-api #297

jordanpadams:
@jimmie can we also be sure to include a check that there is non-zero results returned? basically just making sure there is actually data in the registry too, not just that everything is running.

jimmie:
Sure, we can do that - maybe get a count of documents and include that in the return payload? Note that this will _not_ constitute a failure on the part of the healthcheck (i.e. there may be zero documents but if everything else checks out OK a 200 will still be returned) since the intention for this is to convey to ECS/Fargate of whether the task needs to be recycled.

tloubrieu-jpl:
@jimmie what are you thinking of, regarding the springboot metrics ?

jimmie:
[...]

PR #345, opened 2023-06-10
Introduce new healthcheck/ endpoint for monitoring API health

jimmie:
## 🗒️ Summary
Initial implementation of the healthcheck endpoint.

Attempted to follow the controller/model pattern as much as possible, inserting a new transmuter; SwaggerJavaHealthcheckTransmuter into the inheritance hierarchy of transmuters (just above the terminal SwaggerJavaTransmuter). Healthcheck logic is implemented in the model/HealthcheckLogic class.

Additional tickets will be created to define enhancements to make this endpoint more effective on tracking health in an operations context.

Refs #336

### DIFF
--- a/model/swagger.yml
+++ b/model/swagger.yml
@@ -20,6 +20,8 @@ tags:
     description: search by class of product, bundles, collections, ...
   - name: 4. deprecated
     description: end-points which should not be used anymore
+  - name: 5. healthcheck
+    description: end-point for evaluating system health
 
 externalDocs:
     description: User's Guide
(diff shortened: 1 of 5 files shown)


model/swagger.yml
```diff
@@ -30,6 +32,36 @@ servers:
...
+          content:
+            "*":
+              schema:
+                type: object
+                additionalProperties: true
+        '400':
+          $ref: "#/components/responses/Error"
+        '404':
+          $ref: "#/components/responses/Error"
+        '418':
```
alexdunnjpl, line 54: Is HTTP418 really the way to go here (vs 200?).  The request was successful, so if we have any way to respond to the content of the response JSON (say, looking for the presence of `"healthy": true`), that would be preferable over abusing the status code imho.

service/src/main/java/gov/nasa/pds/api/registry/controller/SwaggerJavaTransmuter.java
```diff
@@ -103,6 +105,29 @@ protected ResponseEntity<Object> processs(EndpointHandler handler, URIParameters
...
 
+  protected ResponseEntity<Map<String,Object>> processHealthcheck() {
+    long begin = System.currentTimeMillis();
+    try {
+      HttpStatus responseStatus = HttpStatus.OK;
+      HealthcheckLogic hcLogic = new HealthcheckLogic(this);
+      Map<String, Object> response = hcLogic.healthcheck();
+
+      if ((boolean) response.get(HealthcheckLogic.FAILURES_PRESENT)) {
+        responseStatus = HttpStatus.I_AM_A_TEAPOT;
```
alexdunnjpl, line 118: Following on from previous comment, if we *can't* do it properly, a comment explaining what is being done here and why would be helpful to $futureDev

alexdunnjpl (changes requested):
Requesting reconsideration of that HTTP418.  LGTM otherwise.

model/swagger.yml
jimmie: Keep in mind that the primary purpose of this endpoint is to inform ECS that the task is not in a good state and should be restarted - the simplest and most straight forward of configuring the associated check in ECS is based on the http status code. I don't want to overload the family of 500's since this to me implies that the ECS workflow it itself not functioning properly.

service/src/main/java/gov/nasa/pds/api/registry/controller/SwaggerJavaTransmuter.java
jimmie: Good point, comment added.

alexdunnjpl (approved):

model/swagger.yml
alexdunnjpl: Roger that - have approved on the basis of the added comment

PR #377, opened 2023-09-14
Fix swagger-ui tags for new /healthcheck and /properties end-points

tloubrieu-jpl:
## 🗒️ Summary
The tags to be removed in ad-hoc code are hardcoded, so I added them.
I also changed the order of the tags so that /healthcheck appears before depracated end-points.


jordanpadams (approved):

jordanpadams:
@tloubrieu-jpl are the failures of the integration tests expected? https://github.com/NASA-PDS/registry-api/actions/runs/6191589159/job/16810114335?pr=377

PR #383, opened 2023-10-25
Optional ccs

tloubrieu-jpl:
## 🗒️ Summary
Adds a configuration to ignore the CCS OpenSearch configuration and only connect to the main registry index.
This is useful when someone want to test on a production opensearch without being authorized to access all the CCS connected openSearch.

Default is CCSEnabled=true so that the current behavior is not altered.

## ⚙️ Test Data and/or Report

One can test with applications.properties section as follow:

"""
# note the port is mandatory even when it is default :80 or :443
openSearch.host=__the_production_opensearch__
openSearch.registryIndex=registry
openSearch.registryRefIndex=registry-refs
openSearch.timeOutSeconds=60
openSearch.CCSEnabled=false
openSearch.username=__your_opensearch_user___
openSearch.password=__your_password__
openSearch.ssl=true
# use only for development purpose, left it to true otherwise
openSearch.sslCertificateCNVerification=true
"""

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/search/OpenSearchConfig.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/search/OpenSearchConfig.java
@@ -38,15 +38,22 @@ public class OpenSearchConfig {
 
   @Value("${openSearch.timeOutSeconds:60}")
   private int timeOutSeconds;
-
+  
   public int getTimeOutSeconds() {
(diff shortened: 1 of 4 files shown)


alexdunnjpl (approved):
consider: replace `CCSEnabled` with `ccsEnabled`, or better yet `crossClusterSearchEnabled`

PR #414, opened 2024-02-28
411 Removes dependency on immature "alternate_ids" product property

### LINKED ISSUE #411: members query return 500 when members do not exist in the registry or alternate_id does not exists
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When I did the request:
https://pds.nasa.gov/api/search/1.0/products/urn:nasa:pds:cassini_iss_cruise:xml_schema::1.0/members

I got a 500 error.

### 🕵️ Expected behavior
I expected to get the members of the collection.  As it does with request https://pds.nasa.gov/api/search/1.0/products/urn:nasa:pds:orex.ovirs:data_calibrated::11.0/members

### 📜 To Reproduce

_No response_

### 🖥 Environment Info

- Version of this software 1.4.0 deployed in production

_No response_

See cloudwatch logs at 2024-02-26T22:31:52.081-05:00:

2024-02-27T03:31:52.079Z ERROR 1 --- [-nio-80-exec-10] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed: java.lang.NullPointerException] with root cause

### 🦄 Related requirements

🦄 #xyz

### ⚙️ Engineering Details

_No response_

alexdunnjpl:
## 🗒️ Summary
`alternate_ids` property was used when resolving collection lidvids for both collection member and bundle member/member queries.

This property is not currently maintained in a way that fulfills its assumed purpose.  This PR removes use of the property, and (because only LIDVIDs are supported in this context) replaces use of string identifiers with `PdsLidVid` objects instead.  If any exceptions are experienced as a result of this change, it indicates an issue elsewhere in the code and exposing that issue is desirable.

## ⚙️ Test Data and/or Report
Tested hits count for both a bundle/member/member query and collection/member query and confirmed same number of hits before/after change.

Tested with both LID and LIDVID identifier in path parameter to confirm that LIDs are correctly resolved to latest-LIDVID as expected.

Tested failing example from #411 to ensure it is now succeeding (with zero hits, as expected, as referenced products are not harvested per @jordanpadams )

## ♻️ Related Issues
Fixes #411

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/LidvidsContext.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/LidvidsContext.java
@@ -3,7 +3,7 @@ package gov.nasa.pds.api.registry;
 import java.util.List;
 
 public interface LidvidsContext {
-  public String getLidVid();
+  public String getProductIdentifierStr();
 
   public Integer getLimit();
 
--- a/service/src/main/java/gov/nasa/pds/api/registry/controller/Member.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/controller/Member.java
@@ -35,7 +35,7 @@ class Member implements EndpointHandler {
       transmuter = ReferencingLogicTransmuter.getBySwaggerGroup(content.getGroup()).impl();
     else
       transmuter = ReferencingLogicTransmuter.getByProductClass(QuickSearch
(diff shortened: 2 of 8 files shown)


tloubrieu-jpl (approved):
That works where it previously failed.

Unfortunately I am not able to understand the code.

PR #477, opened 2024-07-09
Sort search after exceptions

alexdunnjpl:
## 🗒️ Summary
When checking sort and searchAfter value validity, the presence of arguments for both parameters should be considered invalid if they are of non-equal length.


tloubrieu-jpl (approved):
Perfect, thanks @alexdunnjpl

Issue #426, opened 2021-12-23
As a user, by default, I want to search only for the latest versions of all products on the `/products` endpoint


jordanpadams:
## 💪 Motivation
...so that I am not confused by seeing superseded data returned in search results

## 📖 Additional Details

Per the [parent epic](https://github.com/nasa-pds/pds-registry-app/issues/219), and [the design](https://github.com/nasa-pds/pds-registry-app/issues/229) we need to update the API to implement these changes so we can sufficiently understand the version history.

See parent epic for more details.

## ⚖️ Acceptance Criteria
**AC 1:**
**Given** the products `urn:nasa:pds:mars2020.spice::1.0`, `urn:nasa:pds:mars2020.spice::2.0`, and `urn:nasa:pds:mars2020.spice::3.0` have been loaded into the registry (see registry-ref-data) and in `archived` status
**When I perform** a query of the API for `/products` and paginate through the results and/or apply any other filtering that would include this SPICE bundle
**Then I expect** I should get only the latest v3.0 product returned (`urn:nasa:pds:mars2020.spice::3.0`)

**AC 2:**
**Given** the products `urn:nasa:pds:mars2020.spice::1.0`, `urn:nasa:pds:mars2020.spice::2.0`, and `urn:nasa:pds:mars2020.spice::3.0` have been loaded into the registry (see registry-ref-data) and in `archived` status
**When I perform** a query of the API for `/products/urn:nasa:pds:mars2020.spice::1.0`
**Then I expect** I should get the older version 1.0 of the product (`urn:nasa:pds:mars2020.spice::1.0`) because it was explicitly requested

**AC 3:**
_Note: This AC depends upon the completion of this requirement: https://github.com/NASA-PDS/registry-api/issues/427_
**Given** the context products [urn:nasa:pds:context:instrument:crs.vg1::1.0](https://pds.nasa.gov/data/pds4/context-pds4/instrument/vg1.crs_1.0_deprecated.xml) and [urn:nasa:pds:context:instrument:vg1.crs::1.0](https://pds.nasa.gov/data/pds4/context-pds4/instrument/vg1.crs_1.0.xml) ingested into the registry
**When I perform** a query of the API for `/products` and paginate through the results, or perform any additional filtering on the data
[...]

Issue #428, opened 2024-03-28
Implement expected `latest` functionality for all endpoints


jordanpadams:
### 💡 Description

Need to update this for consistency across the API

### ⚔️ Sub-tasks

- https://github.com/NASA-PDS/registry-api/issues/426
- ~#427~
- https://github.com/NASA-PDS/registry-api/issues/503
- ~#504~
- https://github.com/NASA-PDS/registry-api/issues/435
- https://github.com/NASA-PDS/registry-api/issues/436

### Engineering Discussion
I believe we narrowed the `all versions` functionality to only show up when someone specifies `/products/{identifier}/all`, since this is a very specialized use case for trying to track down provenance of a product.

From a user perspective, rarely do end users actually want to search for and/or find old versions of products. They are irrelevant to current science use cases, and we do not want to present that to the users. Newer versions indicate better metadata and better data. This is why the default functionality for all endpoints (except this specialized one) should be `only return the latest versions` (those without a `superseded_by` specified).

This is one of the conversations @tloubrieu-jpl and I were having about functionality in the API that we could potentially abstract away using an OpenSearch query template (or something else)? Since we want this built into all queries of the API by default (except for the very specialized `/products/{identifier}/all`, we should figure out a way to ensure it is always there. For instance, in our legacy solr schema, we have these RequestHandler endpoints that response with pre-filtered results: https://github.com/NASA-PDS/registry-mgr-legacy/blob/main/src/main/resources/collections/data/solrconfig.xml#L817

It looks like OpenSearch/ElasticSearch may use query templates for this?  https://sematext.com/blog/parametrizing-queries-in-solr-and-elasticsearch/

### UPDATE
**Implement this using query parameters**
The new query parameter is "all-versions". the values can be 'true' or 'false'. The parameter is optional, default behavior is all-versions=false. Or actually we could see that the other way around `latest-only=true` by default.

This behavior will replace the suffix `/latest` or  `/all`  to implement the requirement https://github.com/NASA-PDS/registry-api/issues/436 and should be extended to apply to all end-points.

jordanpadams:
📆 05/2024 status: In work. Delayed 1 sprint due to delays in https://github.com/NASA-PDS/registry/issues/185. No impact on build expected

jordanpadams:
@tloubrieu-jpl has this and/or will this be implemented in the improvements you are performing now?

jordanpadams:
[...]

PR #479, opened 2024-07-10
Products endpoint doc update

alexdunnjpl:
## 🗒️ Summary
Trivial swagger doc update - /products only touches the latest-versioned products, yeah @jordanpadams ?

## Applicable requirements
Refs https://github.com/NASA-PDS/registry-api/issues/426

### DIFF
--- a/model/swagger.yml
+++ b/model/swagger.yml
@@ -142,7 +142,7 @@ paths:
       tags:
         - 1. all products
       summary: |
-        search on all PDS data products, including bundles, collections, documentation, context and observational products.
+        search the latest-versioned instances of all PDS data products, including bundles, collections, documentation, context and observational products.
       operationId: product-list
       responses:
         '200':


jordanpadams:
@alexdunnjpl I updated the PR to include the requirement. I also updated the requirement success criteria to include more detailed information. all related requirements regarding versioning should be here in the parent theme:
- https://github.com/NASA-PDS/registry-api/issues/428

jordanpadams:
@alexdunnjpl @tloubrieu-jpl do we think we need separate requirements for each endpoint? or does #426 requirement suffice? In case it is the latter, I will add more success criteria, or I will create some more requirements.

tloubrieu-jpl:
@jordanpadams @alexdunnjpl if we eventually have one test per requirement we should have one requirement per end-point. That would simplify the verification of the coverage of the test. But there might be a smarter way to do that...

tloubrieu-jpl:
And @alexdunnjpl , we want to add an attribute `all-versions` at some point, soon, so we will have to change that documentation again.

tloubrieu-jpl (approved):
Thanks @alexdunnjpl

jordanpadams:
> @jordanpadams @alexdunnjpl if we eventually have one test per requirement we should have one requirement per end-point. That would simplify the verification of the coverage of the test. But there might be a smarter way to do that...

@tloubrieu-jpl +1 will break tout all the requirements.

PR #480, opened 2024-07-10
apply onlyLatest() to /products route

alexdunnjpl:
## 🗒️ Summary
`/products` route should only target latest-versioned products

Applies existing `onlyLatest()` build step to existing route control method


tloubrieu-jpl (approved):
Thanks @alexdunnjpl

PR #482, opened 2024-07-11
Fix the get AWS credential schedule decorators

tloubrieu-jpl:
## 🗒️ Summary
This did not work in the dev deployment on AWS.

## ⚙️ Test Data and/or Report
I tested it locally by having a scheduled task every 3 seconds and it worked. I set it back to every 5 hours in the code.


tloubrieu-jpl:
Was fixed in https://github.com/NASA-PDS/registry-api/pull/513

PR #489, opened 2024-07-11
RegistrySearchRequestBuilder redesign

alexdunnjpl:
## 🗒️ Summary
Revises implementation of RegistrySearchRequestBuilder per earlier discussion with @tloubrieu-jpl 

Our proprietary builder now subclasses `SearchRequest.Builder`, and composes a `BoolQuery.Builder` which is only build when the search request is built.

Documentation added for some functions, but more is needed.

## ⚙️ Test Data and/or Report
I&T tests produce same results vs `develop` - 103 pass, 14 fail

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/search/RegistrySearchRequestBuilder.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/search/RegistrySearchRequestBuilder.java
@@ -52,7 +52,6 @@ public class RegistrySearchRequestBuilder extends SearchRequest.Builder{
         }
       };
 
-
   private ConnectionContext connectionContext;
   private List<String> registryIndices;
   private BoolQuery.Builder queryBuilder;
@@ -105,15 +104,12 @@ public class RegistrySearchRequestBuilder extends SearchRequest.Builder{
     return this.queryBuilder;
   }
 
-  public RegistrySearchRequestBuilder onlyLatest() {
-
-    ExistsQuery supersededByExists = new ExistsQuery.Builder()
-            .field("ops:Provenance/ops:superseded_by")
+  public SearchRequest build() {
+    Query query = this.queryBuilder.build().toQuery();
+    return super
+            .query(query)
+            .trackTotalHits(t -> t.enabled(true))
             .build();
-
-    this.queryBuilder.mustNot(supersededByExists.toQuery());
-
-    return this;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


service/src/main/java/gov/nasa/pds/api/registry/search/RegistrySearchRequestBuilder.java
```diff
@@ -53,222 +52,187 @@ public class RegistrySearchRequestBuilder {
...
 
   }
 
-  public RegistrySearchRequestBuilder sort(List<String> sort) {
+  /**
+   * Implements an alternative to .sort() that accepts strings in API property format.
+   * Currently hardcoded to sort in ascending order only.
+   * @param sortFieldNames
+   */
+  public RegistrySearchRequestBuilder sortFromStrings(List<String> sortFieldNames) {
```
tloubrieu-jpl, line 171: Do we need the fromString in the method name ? It is in the prototype of the function.

tloubrieu-jpl: Now I understand, it conflicts with the sort of the super class.

tloubrieu-jpl: But we might keep the same name for both, did you try ? the polymorphism of java should solve that...

service/src/main/java/gov/nasa/pds/api/registry/search/RegistrySearchRequestBuilder.java
```diff
@@ -53,222 +52,187 @@ public class RegistrySearchRequestBuilder {
...
     return this;
   }
 
 
-  public RegistrySearchRequestBuilder addQParam(String q) throws UnparsableQParamException {
+  /**
+   * Implements an alternative to .fields() that accepts values as strings.
+   * @param fieldNames
+   */
+  public RegistrySearchRequestBuilder fieldsFromStrings(List<String> fieldNames) {
```
tloubrieu-jpl, line 214: same as above,but in this case this is confusing because fieldsFromStrings is not a wrapper of fields but of source ! Which is confusing. Not big deal anyway, we'll find an elegant solution.

tloubrieu-jpl (approved):
That looks very good, thanks @alexdunnjpl .
I am having comments on the naming of functions where you suffixed with FromStrings, but it is not a blocker. That would be nice to have a more elegant, lighter naming but this is not a blocker.

service/src/main/java/gov/nasa/pds/api/registry/search/RegistrySearchRequestBuilder.java
alexdunnjpl: [Necessary, unfortunately](https://jpl.slack.com/archives/C023TQGPW4A/p1720720188360579)

I'd have preferred a polymorphic interface too but ya gets what ya gets

tloubrieu-jpl: That's is too complicated for me to follow the details, but I can feel that there is some background here. Thanks !

service/src/main/java/gov/nasa/pds/api/registry/search/RegistrySearchRequestBuilder.java
alexdunnjpl: The rationale for FromStrings here is for consistency with the other *FromStrings methods - they're all doing basically the same thing, abstracting away the verbose construction of proper classed objects from Strings

If I were writing it from scratch, I'd have named it `sourceIncludesFromStrings()` instead, since it's setting the _source _includes, as you say, but the method was already there and I assumed it had been named `fields()` because you found it more intuitive.  If not, I'll change it to `sourceIncludesFromStrings()` if you agree that's an improvement?

service/src/main/java/gov/nasa/pds/api/registry/search/RegistrySearchRequestBuilder.java
alexdunnjpl: The short version is "Java doesn't support polymorphism where the only difference is in the type `T` of some parameter generic  `SomeType<T>`

So `List<String>` and `List<PdsProductIdentifier>` are the same type from Java's perspective, when checking for conflicts.

alexdunnjpl:
Merging on the basis of approval, but let me know what you think about https://github.com/NASA-PDS/registry-api/pull/489#discussion_r1674807084 and I'll resolve that.

PR #517, opened 2024-07-30
Implement common multiple-products build method

### LINKED ISSUE #515: Refactor common RegistrySearchRequestBuilder dot-chains into applyMultipleProductsDefaults()
### 💡 Description

Should be relatively straightforward, and @jordanpadams this will ensure consistent application of stuff like pagination, querystring constraints, latest-only default, etc, across endpoints without the need for a million requirements/tests like we've seen with latest-only recently

### ⚔️ Parent Epic / Related Tickets

_No response_

alexdunnjpl:
## 🗒️ Summary
Rolls build options common to all routes returning multiple product hits into a common execution path.

## ⚙️ Test Data and/or Report
130 pass, 10 fail, before and after changes

## ♻️ Related Issues
fixes #515

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/controllers/ProductsController.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/controllers/ProductsController.java
@@ -249,11 +249,7 @@ public class ProductsController implements ProductsApi, ClassesApi {
       Integer limit, String q, List<String> sort, List<String> searchAfter) throws Exception {
 
     SearchRequest searchRequest = new RegistrySearchRequestBuilder(this.connectionContext)
-            .constrainByQueryString(q)
-            .addKeywordsParam(keywords)
-            .fieldsFromStrings(fields)
-            .paginate(limit, sort, searchAfter)
-            .onlyLatest()
+            .applyMultipleProductsDefaults(fields, q, keywords, limit, sort, searchAfter, true)
             .build();
 
     SearchResponse<HashMap> searchResponse =
@@ -437,9 +433,7 @@ public class ProductsController implements ProductsApi, ClassesApi {
       }
 
       SearchRequest searchRequest = searchRequestBuilder
-              .fieldsFromStrings(fields)
-              .paginate(limit, sort, searchAfter)
-              .onlyLatest()
+              .applyMultipleProductsDefaults(fields, "", List.of(), limit, sort, searchAfter, true)
               .build();
 
       SearchResponse<HashMap> searchResponse =
@@ -449,7 +443,7 @@ public class ProductsController implements ProductsApi, ClassesApi {
 
       return formatMultipleProducts(products, fields);
 
-    } catch (IOException | OpenSearchException e) {
+    } catch (IOException | OpenSearchException | UnparsableQParamException e) {
       throw new UnhandledException(e);
     }
   }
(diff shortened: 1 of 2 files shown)


alexdunnjpl:
Some routes do not currently have qparams/implementations for querystring and keywords, in which case no-op placeholders (null-string, empty keyword list) have been used.

Those constraints may be desirable for those routes, in which case qparams can be defined and the no-op placeholders replaced with actual values from the request.

alexdunnjpl:
Re-running branch tests, as failure was not replicable locally

alexdunnjpl:
Not really sure where to go from here...

@tloubrieu-jpl any ideas?

image

tloubrieu-jpl:
Hi @alexdunnjpl , is the latest case you raised an issue ? I am running now the test against the develop branch and all succeeded except one related to the legacy_registry.

service/src/main/java/gov/nasa/pds/api/registry/controllers/ProductsController.java
```diff
@@ -367,7 +363,7 @@ private PdsLidVid resolveLatestLidvid(PdsProductIdentifier identifier)
     SearchRequest searchRequest = new RegistrySearchRequestBuilder(this.connectionContext)
             .matchLid(identifier.getLid())
             .fieldsFromStrings(List.of())
-            .onlyLatest()
+            .noSupersededProducts()
```
tloubrieu-jpl: I feel like positive assertion are easier to understand, but I see your idea of being more explicit regarding what this method actually does. No need to revert that.

service/src/main/java/gov/nasa/pds/api/registry/search/RegistrySearchRequestBuilder.java
```diff
@@ -106,6 +106,39 @@ public BoolQuery.Builder getQueryBuilder() {
...
+          List<String> searchAfterFieldValues,
+          Boolean noSupersededProducts
+  ) throws UnparsableQParamException, SortSearchAfterMismatchException {
+    this
+      .fieldsFromStrings(includeFieldNames)
+      .constrainByQueryString(queryString)
+      .addKeywordsParam(keywords)
+      .paginate(pageSize, sortFieldNames, searchAfterFieldValues);
+
+    if (noSupersededProducts) {
```
tloubrieu-jpl: I would not have had this in the default since it is not default anymore as sometimes we have it and sometimes we don't but this is very minor.

tloubrieu-jpl (approved):
All the tests work on my side.

service/src/main/java/gov/nasa/pds/api/registry/controllers/ProductsController.java
alexdunnjpl: @tloubrieu-jpl I tend to agree. `excludeSupersededProducts()`?

service/src/main/java/gov/nasa/pds/api/registry/search/RegistrySearchRequestBuilder.java
alexdunnjpl: the no-superseded filter?

I thought it was considered default, but overridable by a yet-to-be-implemented qparam?

alexdunnjpl:
> Hi @alexdunnjpl , is the latest case you raised an issue ? I am running now the test against the develop branch and all succeeded except one related to the legacy_registry.

This is the failure I'm referring to - for some reason when the tests are run in the action, it fails. I don't see any potential for the test definitions to be different between the runner and my local, since the URL is given in the action test output...

alexdunnjpl:
@tloubrieu-jpl is the deep-archive compatibility check new-ish?  Is it a blocker that it's failing?

alexdunnjpl:
Merging on the basis of approval

PR #529, opened 2024-08-21
Handle Accept header with multiple values

tloubrieu-jpl:
## 🗒️ Summary

We fix the handling of multiple Accept header as those provided by web browser.

Since the handling of this header value got a bit more complex, we moved the related code to the MVC configuration specifically.

## ⚙️ Test Data and/or Report
* Reference to regression test included in code (preferred wherever reasonable)
* Attach test data here + outputs of tests

Use PR in registry repository https://github.com/NASA-PDS/registry/pull/318

Unit tests have also been added.

## ♻️ Related Issues

Fixes requirement #439

### DIFF
--- a/service/src/main/java/gov/nasa/pds/api/registry/configuration/WebMVCConfig.java
+++ b/service/src/main/java/gov/nasa/pds/api/registry/configuration/WebMVCConfig.java
@@ -1,6 +1,8 @@
 package gov.nasa.pds.api.registry.configuration;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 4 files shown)


tloubrieu-jpl:
Hi @jordanpadams , you noticed that bug yesterday and I cowardly pretended that was not a bug ;)

Actually, today I wanted to fix that because I was thinking if the API URL will not work from a browser that will make us have to answer more question and not look too good. So I looked at that, thinking this was more like a configuration issue, but realized there was an actual bug when multiple formats are proposed as supported returned format in the Accept header, e.g. text/html or application/svg, .... This is what web browsers do.
This is now fixed.

alexdunnjpl (approved):
LGTM

jordanpadams:
@tloubrieu-jpl I notice this has not yet been deployed to production? Unfortunately, most folks do the same thing I do which is put a URL into a browser, which doesn't work :-/ .

PR #533, opened 2024-09-05
Test re-enabling deep archive

jordanpadams:


alexdunnjpl (approved):
I'm confused that it _was_ disabled and is now re-enabled and working, but... as long as it's working!

PR #639, opened 2025-04-22
Terms faceting implementation

### LINKED ISSUE #283: As a client developer, I want to facet on 1 or more fields in the registry
## 💪 Motivation
...so that I can enable faceting for my client across one or more fields to help guide a user towards their expected search results.

## ⚖️ Acceptance Criteria
**Given** the PDS API online documentation
**When I perform** I know how to perform a facetting search on the API and when I perform it
**Then I expect** to get the unique values of the fcetted fields and the number of occurance of these values for the current search criteria

## ⚙️ Engineering Details

Tightly coupled to NASA-PDS/registry-api#284

alexdunnjpl:
## 🗒️ Summary
Implements terms faceting (only - not range/histogram faceting of non-terms values)

## Outstanding considerations
- @jordanpadams @tloubrieu-jpl the `facet-limit` qparam is relatively rigid - can we confirm that users and @anilnatha @eddiesarevalo will have no desire to set different facet bucket-size values per-facet, before cementing this as the interface?

- I've defaulted the value of `facet-limit` to `10` rather than the `100` specified in the slide deck examples - this is (technically) more performant, and for search-ui faceting it's a more practical default.

## ⚙️ Test Data and/or Report
New test implementation pending

## ♻️ Related Issues
Resolves #283

### DIFF
--- a/model/swagger.yml
+++ b/model/swagger.yml
@@ -162,6 +162,7 @@ paths:
       - $ref: "#/components/parameters/Query"
       - $ref: "#/components/parameters/Sort"
       - $ref: "#/components/parameters/SearchAfter"
+      - $ref: "#/components/parameters/FacetFields"
 
   /products/{identifier}:
     get:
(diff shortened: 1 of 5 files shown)


tloubrieu-jpl (approved):
Thanks @alexdunnjpl , I will merge so that I can integrate to my own feature branch.

PR #637, opened 2025-04-16
reintroduce field processing

al-niessner:
## 🗒️ Summary
Reinserted field processing once I figured out where it went (turns out there was nice comment suggesting where it should go). Used the min/max fields for the output type to constrain and/or fields as necessary to the request.

## ⚙️ Test Data and/or Report
All automated tests below should pass.

## ♻️ Related Issues
Closes #548

PR #643, opened 2025-05-01
Re-engineer the API response transformations depending on the Accept headers

### LINKED ISSUE #548: Responses do not return correct metadata when `fields` parameter is specified
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When I did used the `fields` query parameter, I noticed the API returned many other fields I did not ask for

### 🕵️ Expected behavior

I expected only the fields I asked for to be returned

### 📜 To Reproduce

```
curl -L --get https://pds.nasa.gov/api/search/1/products/urn:nasa:pds:msl_gt_diagenesis_supplement?fields=lid --header 'Accept: text/csv'
```

Confirmed this also does not work for:

* `application/kvp+json`
* `application/json`
* `application/xml`

### 🖥 Environment Info

Mac OSx

### 🦄 Related requirements

🦄 #459 
🦄 #445

### ⚙️ Engineering Details

_No response_

### 🎉 Integration & Test

_No response_

tloubrieu-jpl:
## 🗒️ Summary
Re-design of the formally named product business logic, now called transformer.

I made the change when I realize the initial fix for #548 proposed in PR #637 required more in-depth revision of the code to have a readable/maintainable code.

This PR also concentrates all the fields constant definitions in the PdsProperty class.

## ⚙️ Test Data and/or Report
Use the docker compose setup. postman suite successful locally, missed one test in the github CICD because of PR https://github.com/NASA-PDS/registry-sweepers/pull/167 not merged yet.

## ♻️ Related Issues

Fixes #548


tloubrieu-jpl:
Note one remaining error related to the PR https://github.com/NASA-PDS/registry-sweepers/pull/167

service/src/main/java/gov/nasa/pds/api/registry/model/transformers/PdsProductTransformer.java
```diff
@@ -0,0 +1,92 @@
...
+    log.debug("transform: fields: {}, excluded fields: {}", fields, EXCLUDED_PROPERTIES);
+    PdsProducts products = new PdsProducts();
+    Set<String> uniqueProperties = new TreeSet<String>();
+    Map<String, List<String>> properties;
+
+    for (Map<String, Object> kvp : input.getProducts()) {
+      try {
+        PdsProduct product = SearchUtil.entityProductToAPIProduct(objectMapper.convertValue(kvp, EntityProduct.class), this.baseURL);
+        
+        // TODO check why every value is a String
```
alexdunnjpl, line 64: Is this TODO prior to merge?

service/src/main/java/gov/nasa/pds/api/registry/model/transformers/PdsProductTransformer.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @Override
+  @LogExecutionTime
+  public Object transform(Map<String, Object> kvp, List<String> fields) {
+    EntityProduct ep = objectMapper.convertValue(kvp, EntityProduct.class);
+    PdsProduct product = SearchUtil.entityProductToAPIProduct(ep, this.baseURL);
+    
+    Map<String, List<String>> properties = (Map<String, List<String>>) getFilteredProperties(kvp, fields, EXCLUDED_PROPERTIES);
+    // TODO: findout why the getFilteredProperties method is used here. Should we add fields as a
```
alexdunnjpl: Is this TODO prior to merge?

service/src/main/java/gov/nasa/pds/api/registry/model/properties/PdsProperty.java
alexdunnjpl, line 1: [prefer] these constant static members are good/needed, but imho these members should return instances of `PdsProperty`, rather than their underlying string values.

alexdunnjpl (changes requested):
Looks good, though now would be a good time to introduce summary documentation explaining (concisely) what the new classes/constructs are intended to do, for the benefit of anyone skimming the code.

The only other thing I'd suggest is the `PdsProperty` constants thing (see granular comment)

Did you take a look at the new (automated) code quality issues?  Don't need to fix them, provided you looked at them and addressed any which were worthwhile.

service/src/main/java/gov/nasa/pds/api/registry/model/transformers/PdsProductTransformer.java
tloubrieu-jpl: I added the TODO while looking at the code. It was not there before but the line on which I am commenting was there before. I did not take the time to confirm if this has an impact on the responses of the API. 

That sounds critical may be I should...

service/src/main/java/gov/nasa/pds/api/registry/model/transformers/PdsProductTransformer.java
tloubrieu-jpl: This one I can remove I believe.

service/src/main/java/gov/nasa/pds/api/registry/model/properties/PdsProperty.java
tloubrieu-jpl: I thought that was a good idea to and I wanted to do that initially but when starting that I realized half way how heavy that would be in the code to "contruct" a PdsProperty for each field that we use.

I looked if we can overload a String type with the behavior we need to convert opensearch fields to/from api fields but this is not possible in Java to overload a String object.

I am open to your ideas on that , I might have missed an easier/friendlier option.

service/src/main/java/gov/nasa/pds/api/registry/model/transformers/PdsProductTransformer.java
tloubrieu-jpl: @alexdunnjpl I checked and all the numbers are converted to string in production and with this version of the code. I will create a different ticket to work on that.

tloubrieu-jpl:
> Looks good, though now would be a good time to introduce summary documentation explaining (concisely) what the new classes/constructs are intended to do, for the benefit of anyone skimming the code.

I wanted to explore the option of having more javadoc to avoid to maintain documentation on the side, but that did not sound like a pragmatic approach for a Friday evening.

I will also create a separate ticket to create this documentation, whatever the format we decide for it.

service/src/main/java/gov/nasa/pds/api/registry/model/properties/PdsProperty.java
alexdunnjpl: (all imho, and assuming I understand "heavy" here)

Lightweight/brief is no good if it comes at the cost of implementing the desired behaviour.  In this case, treating fields as strings is a shortcut, and it's one that has bitten us in the past.  Any time saved not typing `new PdsProperty()` is guaranteed to be eaten up the ~first~ next time a dev forgets whether they're dealing with an OpenSearch or API string and fails to invoke the necessary conversion, then has to debug it.

The introduction of the `PdsProperty` construct (and incremental conversion of API internals to use it) is fundamental OOP imho, and as for brevity... it's _Java_.

tloubrieu-jpl: Thanks @alexdunnjpl , I am getting your point. I asked chatGPT which does not propose alternatives to naked String or PDsProperty object.
Do you mind if we have a specific must-have ticket for this re-engineering ? I am proposing that because at least, in this ticket we have been able to concentrate the constants in a single place. Making all of them PdsProperty is a significant re-engineering process since this constants are used everywhere and the fields as well. I am trying to release a stable version of the registry-api for the build 15.1 and this PR is blocking it.

tloubrieu-jpl: I've created this ticket https://github.com/NASA-PDS/registry-api/issues/647

alexdunnjpl: @tloubrieu-jpl If the issue is that the newly-defined constants are referenced a lot in the new content in this PR, it seems like it should be a quick/simple enough thing to search/replace all references to the current `String` constants `PdsProperty.SOMECONSTANT` -> `PdsProperty.SOMECONSTANT.value`prior to updating the constants to return `PdsProperty`.

That way, the construct itself can be defined "properly" even if the references to it are only using the underlying `String` value.  This way, any work done between now and when the rework ticket is eventually completed will be able to use the "proper" constants and won't add work to the rework ticket.

That's how I was able to make incremental changes throughout the codebase to use `PdsProperty` in the past, avoiding the (as you noted - significant) work of overhauling everything all at once.

That would be my strong suggestion, but if you insist I can approve as-is.

tloubrieu-jpl: Ok I see I ll try to complete that this week.

alexdunnjpl: 👍 👍 👍

alexdunnjpl (approved):
Lookin' goood

tloubrieu-jpl:
The failed test in github action is related to sweeper not working in the registry integration, either because of a delayed update oof the sweeper docker image or an actual bug. No issues with the API itself.

PR #664, opened 2025-06-30
Delete lexer/CHANGELOG.md

jordanpadams:
## 🗒️ Summary
Cleanup Remnants from when this was made up of several separate repos


jordanpadams:
This is breaking stuff so just going to leave the trash on the pile :-)

PR #671, opened 2025-07-08
Add missing mandatory file for application/json and application/xml response format

tloubrieu-jpl:
## 🗒️ Summary
Bug has been noticed when validating pds-api-client on API 1.6.0 in production.

Tested locally with the docker compose setup.

This PR should also be merged in the develop branch.


tloubrieu-jpl:
3 errors in automated tests are ok for now:
- one related to the number of properties which lately changed due to the publication of the JSON LDD which was previously missing.
- 2 related to the legcy registry synchronization which is know to not work.

PR #700, opened 2025-10-23
Add API support for checking if a search field exists in a result (e.g. not null)

### LINKED ISSUE #406: As a user, I want to query for documents where a specific search field exists in the document
### Checked for duplicates

Yes - I've already checked

### 🧑‍🔬 User Persona(s)

Data User

### 💪 Motivation

...so that I can find products that contain (or do not contain) a specific field value in the document.

### 📖 Additional Details

The `exists` operator allows users to query for documents based on the presence (or absence) of specific fields in the document.

> **⚠️ Syntax Update:** The original implementation used **postfix** notation (e.g., `field exists`). Based on API Working Group feedback, the syntax has been updated to **prefix** notation (e.g., `exists field`) in #727 for more intuitive usage.

---

#### Original Implementation (Postfix - Deprecated)

**Syntax:**
- Check if exact field exists: `pds:Target/pds:name exists`
- Check if field does NOT exist: `not (pds:Investigation/pds:stop_date_time exists)`
- Regex pattern matching: `"pds:Target.*" exists` - matches any field starting with `pds:Target`
- Complex regex patterns: `".*Bounding_Coordinates.*" exists`

---

#### Updated Syntax (Prefix - See #727)

**Syntax:**
- Check if exact field exists: `(exists pds:Target/pds:name)`
- Check if field does NOT exist: `(not exists pds:Investigation/pds:stop_date_time)`
- Regex pattern matching: `(exists "pds:Target.*")` - matches any field starting with `pds:Target`
- Complex regex patterns: `(exists ".*Bounding_Coordinates.*")`

---

**Implementation Notes:**
- For exact field names (unquoted), the query checks for that specific field in the document
- For quoted strings, the value is treated as a Java regex pattern and matched against all known field names from the OpenSearch mapping
- When using regex patterns, all matching field names are retrieved and an existence check is created for each
- Returns error if regex pattern matches no known fields
- `exists` returns true only if the field is present AND has a non-null/non-empty value
- For multi-valued fields, `exists` returns true if at least one value is present

**Use Case Examples (Updated Syntax):**
* Find all products that have bounding box coordinates: `(exists cart:Bounding_Coordinates/cart:north_bounding_coordinate)`
* Find investigations that are still ongoing (no stop date): `(not exists pds:Investigation/pds:stop_date_time)`
* Find products with any target field using regex: `(exists "pds:Target.*")`
* Find products missing author information: `(not exists pds:Citation_Information/pds:author_list)`
* Complex pattern matching: `(exists ".*Coordinates.*")` finds all fields containing "Coordinates"

Related: See #402 for related query operator issues.
Related: See #727 for the syntax update from postfix to prefix notation.

### Acceptance Criteria

**Given** I am querying the registry API
**When** I use `(exists field_name)` with an exact field name in my query
**Then** the API returns only documents where the specified field exists and has a non-null/non-empty value

**Given** I am querying the registry API
**When** I use `(not exists field_name)` in my query
**Then** the API returns only documents where the specified field does not exist or is null/empty

**Given** I am querying for a multi-valued field
**When** I use `(exists field_name)` and at least one value is present
**Then** the API returns that document in the results

**Given** I want to check for fields matching a regex pattern
**When** I use `(exists "regex_pattern")` with a quoted string
**Then** the API:
- Retrieves all field names from the OpenSearch mapping
- Matches the regex against all known field names
- Returns documents where any matched field exists with a non-null/non-empty value
- Returns an error if the regex matches no known field names

**Given** I combine `exists` with other query operators
**When** I use queries like `(exists pds:Target/pds:name) and pds:Target/pds:type eq "Planet"`
**Then** the API correctly applies both conditions and returns matching documents

### ⚙️ Engineering Details

Implementation completed in PR #700 (postfix syntax):
- Updated ANTLR4 lexer grammar (`Search.g4`) to support postfix `EXISTS` keyword with both FIELD and STRINGVAL tokens
- Updated `Antlr4SearchListener` to handle existence checks:
  - Exact field name matching for unquoted field names
  - Java regex pattern matching for quoted strings
  - Retrieves mapping from OpenSearch via `ProductsController.productPropertiesList()` for regex matching
  - Generates OpenSearch `ExistsQuery` for each matched field
- Proper handling of NOT for negation
- Error handling when regex matches no fields

**Syntax Update:** See #727 for the update from postfix to prefix notation based on API Working Group feedback.

al-niessner:
## 🗒️ Summary
Allows queries to collect documents based on the existence of a field.

## ⚙️ Test Data and/or Report
Do not know yet

## ♻️ Related Issues
Closes #406


al-niessner:
@tloubrieu-jpl 

The branch test failed because of registry-loader. Really not part of these code changes...

al-niessner:
@jordanpadams @tloubrieu-jpl 

Let me know about the bomb out. If you want it then I will add the statement to do it. Otherwise it is ready; meaning it fails silently  from the end users perspective.

jordanpadams:
@al-niessner per my comment on the issue, let's return an error if something isn't as expected unless @tloubrieu-jpl has other ideas.

jordanpadams:
@al-niessner can we had a 2+ regression tests to the postman suite to support this?

jordanpadams (changes requested):
See other comments for error handling and tests needed

service/src/main/java/gov/nasa/pds/api/registry/model/Antlr4SearchListener.java
```diff
@@ -178,6 +177,44 @@ else if (this.operator == operation.lt)
 
   }
 
+  @Override
+  public void exitExistence(SearchParser.ExistenceContext ctx) {
+    ArrayList<Query> checks = new ArrayList<Query>();
+    final String fieldName = SearchUtil.jsonPropertyToOpenProperty(ctx.FIELD().getSymbol().getText());
+    final String regexp = ctx.STRINGVAL().getText();
```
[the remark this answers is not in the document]
al-niessner: False positive. Because the antlr language requires one or the other they both cannot be null.

service/src/main/java/gov/nasa/pds/api/registry/model/Antlr4SearchListener.java
```diff
@@ -178,6 +177,44 @@ else if (this.operator == operation.lt)
...
+    String theKey = "''";
+    
+    if (fieldName != null && !fieldName.isBlank()) {
+      theKey = fieldName;
+      checks.add(new ExistsQuery.Builder().field(fieldName).build().toQuery());
+    } else if (regexp != null && !regexp.isBlank()) {
+      theKey = regexp;
+      if (knownFieldNames.isEmpty()) {
+        try {
+          for (PropertiesListInner property : ProductsController.productPropertiesList(this.connectionContext).getBody()) {
```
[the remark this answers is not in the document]
al-niessner: False positive. Checking if a list is empty is cheap compared to downloading. Also, once downloaded do not need to do it many times if q= contains more than one exists. Lastly, iterating over an empty collections always does the correct thing. All in all, this suggestion has to be the worst.

service/src/test/java/gov/nasa/pds/api/registry/opensearch/Antlr4SearchListenerTest.java
```diff
@@ -189,7 +189,7 @@ void testEnterOrStatement() {
   void testExitOrStatement() {
     listener.enterOrStatement(Mockito.mock(SearchParser.OrStatementContext.class));
     listener.exitOrStatement(Mockito.mock(SearchParser.OrStatementContext.class));
-    assertTrue(listener.getBoolQuery().minimumShouldMatch() == "1",
+    assertSame("1", listener.getBoolQuery().minimumShouldMatch(),
```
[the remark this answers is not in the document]
al-niessner: Funny, sonarque wanted assertSame.

al-niessner:
> @al-niessner can we had a 2+ regression tests to the postman suite to support this?

@jordanpadams @tloubrieu-jpl how?

jordanpadams:
@al-niessner just created some documentation for this here: https://github.com/NASA-PDS/registry/pull/445

I will also email the internal wiki page.

jordanpadams:
@al-niessner can you take a look at this merge conflict?

jordanpadams:
Also it looks like the branch testing is failing

al-niessner:
@jordanpadams 

Sorry, those conflicts were to reduce issues reported by sonarcube. They are fairly cosmetic but reduce the over complaints by sonarcube.

al-niessner:
@jordanpadams 

Looks like you still have a problem with looking at an old lexer or something when trying to build the test.

jordanpadams:
@tloubrieu-jpl can you please review and merge or resolve the branch testing issues?

lexer/src/test/java/api/pds/nasa/gov/api_search_query_lexer/TestParsing.java
```diff
@@ -113,14 +123,220 @@ void testTemporalRange() {
...
+
+  @Override
+  public void enterOperator(OperatorContext ctx) {
+    // TODO Auto-generated method stub
+
+  }
 
+  @Override
+  public void exitOperator(OperatorContext ctx) {
+    // TODO Auto-generated method stub
```
tloubrieu-jpl: Minor comment, but I think we could replace those TODO comments with "Nothing needs to be done here."

lexer/src/main/antlr4/gov/nasa/pds/api/registry/lexer/Search.g4
```diff
@@ -1,8 +1,9 @@
 grammar Search;
 
 query : queryTerm EOF ; 
-queryTerm : comparison | likeComparison | group ;
+queryTerm : comparison | likeComparison | existence | group ;
 group : NOT? LPAREN expression RPAREN ;
+existence : ( FIELD | STRINGVAL ) EXISTS ;
```
tloubrieu-jpl, line 6: I understand we want to be flexible, but I would only consider FIELD so to be more readable/simple on how the syntax is handled.
I mean it is always a field expected here, either it can have "" around of not is another topic that we don't need to spend time on now.

lexer/src/test/java/api/pds/nasa/gov/api_search_query_lexer/TestParsing.java
tloubrieu-jpl: Less minor, when I ran maven install it gave me errors for those lines:
```
[ERROR] /Users/loubrieu/git/registry-api/lexer/src/test/java/api/pds/nasa/gov/api_search_query_lexer/TestParsing.java:[70,34] cannot find symbol
[ERROR]   symbol: variable field
[ERROR] /Users/loubrieu/git/registry-api/lexer/src/test/java/api/pds/nasa/gov/api_search_query_lexer/TestParsing.java:[71,40] cannot find symbol
[ERROR]   symbol: variable field
[ERROR] /Users/loubrieu/git/registry-api/lexer/src/test/java/api/pds/nasa/gov/api_search_query_lexer/TestParsing.java:[73,34] cannot find symbol
[ERROR]   symbol: variable number
[ERROR] /Users/loubrieu/git/registry-api/lexer/src/test/java/api/pds/nasa/gov/api_search_query_lexer/TestParsing.java:[74,41] cannot find symbol
[ERR
```

tloubrieu-jpl:
@al-niessner This is great to see tests in postman, it is very easy to review that the code works as expected this way. I like it.

One detail we could discuss:

Case A: For the queries `/products?q=(".*illogical_identifier" exists)` and `/products?q=not (".*illogical_identifier" exists)` the returned status is 400 BAD REQUEST with message: `cannot match any field names to .*illogical_identifier`
That is not bad but, if the field cannot be found we could also return no products or all the products.

Case B: That is a similar basic behavior that you are having on query `/products?q=(pds:Identification_Area.pds:illogical_identifier exists)` which returns 0 results, and I am finding it more intuitive.

Ideally, and to be consistent in both cases A and B, I would have a result (everything or empty) with HTTP status 200.

To me what seems obvious is: 

**Given:** a field or filed pattern is not available in any product
**When:**  the user asks for the products for which this field exists, 
**Then**: it returns nothing. 

**Given:** a field or filed pattern is not available in any product
**When:**  the user asks for the products for which this field does not exists, 
**Then:** it returns everything. 

Without a need to expose errors to the user.

I'll leave it to @jordanpadams .

al-niessner:
@tloubrieu-jpl @jordanpadams 

Do we always return a status of 200 but a message in the result that says "oh, by the way I really failed because you gave me crud to work with". I do not think so. We return a bad http status.

Wildcarding a field to something that does not exist should be an error. Sure, make the message better or whatever, but let the user know up front without ambiguity that they should not have expected anything to work as they thought it should and without ambiguity why. 

If we always returned a good/bad item in our results that users had to always check instead of using http errors, then sure. We do not - well did not. Suggesting now that the user just needs to look at the http return status except in some cases is going to cause you sooo much trouble. It is worse than i before e except after c but for height, weird, etc.

tloubrieu-jpl:
Yes... I am ok either way. I was leaning toward the simple systematic approach. The example I had in mind is a field which is valid (in the PDS4 model) but has never been used.

Searching for simple, with less side cases, is the way my mind is wired, that does not mean others and especially users are wired the same.

At some point in my edition of my previous comment I added the option to add warnings in the response's summary to tell the users about that, by the way, the field that you are asking for does not exist in any product, but eventually I realized that it is redundant with the exists operator itself: 
Question: "return the products for which the field exists"
Response: "if the fields exists: we return products, if does not exist we don't return products.".

I suggest to let the week-end pass over it and come back to it next week at a break out or next year. I am fine to merge the PR as-is for now anyway.

tloubrieu-jpl (approved):
Perfect,.

lexer/src/main/antlr4/gov/nasa/pds/api/registry/lexer/Search.g4
tloubrieu-jpl: We can ignore this comment, StringVal is used for REGEX. I did not get it first.

tloubrieu-jpl:
Follow up ticket https://github.com/NASA-PDS/registry-api/issues/712 (optional)

lexer/src/test/java/api/pds/nasa/gov/api_search_query_lexer/TestParsing.java
al-niessner: No opinion.

lexer/src/main/antlr4/gov/nasa/pds/api/registry/lexer/Search.g4
al-niessner: No, because you cannot easily tell if something is a regex or not. Mostly because regex are complex -> is a:b:c.d a literal or regex. So, use the quotes for that good old fashion "without ambiguity" that I keep spouting.

If I can get the globing to work, then the regex here will become globing and not need quotes.

al-niessner:
@tloubrieu-jpl 

I think you are fully covered:
1. field exists in LDD and some documents -> non-zero list of returned
2. field exists in LDD and no documents -> empty list returned
3. field does not exist in LDD -> error

Remember the field name error is based on loaded LDDs not fields used in the loaded documents. It does bring up the interesting question of last LDD loaded was v24 and a user is using a field from LDD v26. They will get the error from (3) but should they? I agree with you that it might be worth discussing.

jordanpadams:
@al-niessner @tloubrieu-jpl i am in line with the discussion above and agree if the field does not match, we should throw and error, versus return 0, since in the latter case you wouldn’t know the difference between a valid query with 0 results and an invalid query.

for the attributes that do not yet exist in the data, but do in the IM, this may be some sort of secondary query to the index containing the LDDs to see if it exists at all before return the error. Not sure how feasible that is, but eventually we want to provide that kind of search functionality for attributes in general.

jordanpadams:
Since from the conversations above it appears that the tests failing is OK, I am merging this PR

alexdunnjpl:
@jordanpadams @al-niessner @tloubrieu-jpl coming in way (too?) late here, but I'm not sure 400 is the way to go.

If you make a request for documents where `someValidField` exists/notexists, but where no documents have *yet* been loaded with `someValidField` (ergo it isn't in the mappings *yet*), that doesn't make it a bad request, nor does it make `someValidField` a garbage value.

I'd also assert that querying "`someGarbageField` not exists" is not a bad request, just a pointless one.

I'd strongly urge treating this the same way OpenSearch does, i.e. "notexists" -> no attribute or non-null, "exists" -> non-null value.

I promise this will cause confusion later if left as-is, especially if/when we want to implement a thin interface for OpenSearch queries, as has been discussed recently.

jordanpadams:
@alexdunnjpl thanks for the input. I will update the requirement

al-niessner:
@alexdunnjpl 

It will cause confusion later too when the user is absolutely certain their wildcard should when it simply maps to no known or existing field names. It is better to signal that the wildard cannot match anything known then reply with an empty list that no documents have been found.

The problem is you are using a "valid" result for error conditions. An empty list is allowed. Asking for field names that do not exist and cannot exist by nature of the LDD is strictly an error not a valid empty list.

I do agree that all solutions will cause confusion tomorrow. Just pick your pain. I chose the pain of complaints about an error rather than repeat explaining after hours of checking the wildcard that it cannot result in a valid field name.

alexdunnjpl:
@al-niessner I may be lacking context/detail coming in cold - are we talking about wildcards which _don't_ resolve to anything extant, or which _can't_ resolve to anything conceivably valid?

In the latter case I'd agree it's a bad request, if the former then I don't think it's on us to assume that a user is always certain their wildcard expression will return nonzero matching expansions, and treat the wildcard like an invalid enum value.

But I'm speaking from the armchair - as long as it's been considered I trust y'all.

PR #718, opened 2026-01-06
document behaviour

### LINKED ISSUE #456: Some properties return "null" values
### Checked for duplicates

No - I haven't checked

### 🐛 Describe the bug

When I did the query https://pds.mcp.nasa.gov/api/search/1/products/urn:nasa:pds:msl_chemcam_psv_calibrated:data_rad

`ops:Provenance.ops:superseded_by` and `ps:Provenance.ops:registry_sweepers_provenance_version` fields returned "null" values as strings

### 🕵️ Expected behavior

I expected not to see those fields at all or they would not have null values.

### 📜 To Reproduce

Make the query provided above.

### 🖥 Environment Info

Production deployment

### ⚙️ Engineering Details

That could be caused by registry-sweepers.

### 🎉 Integration & Test

_No response_

alexdunnjpl:
fixes https://github.com/NASA-PDS/registry/issues/456

### DIFF
--- a/model/swagger.yml
+++ b/model/swagger.yml
@@ -2,13 +2,12 @@ openapi: 3.0.0
 info:
   description: |
     Registry API enabling advanced search on PDS data and metadata. The API provides end-points to search for bundles, collections and any PDS products with advanced search queries. It also enables to browse the archive hierarchically downward (e.g. collection/s products) or upward (e.g. bundles containing a product).
-    
     Property values are cast to string in responses due to limitations of JSON typing, and should be interpreted by the client/user according to the data dictionary.
     As a result of the string cast, some missing values might be found as "null".
-  version: 1.3.0
   title: PDS Registry Search API
   termsOfService: 'http://pds.nasa.gov'
   contact:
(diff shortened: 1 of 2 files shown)


alexdunnjpl:
Branch test failure unrelated to PR

model/swagger.yml
```diff
@@ -2,6 +2,8 @@ openapi: 3.0.0
 info:
   description: |
     Registry API enabling advanced search on PDS data and metadata. The API provides end-points to search for bundles, collections and any PDS products with advanced search queries. It also enables to browse the archive hierarchically downward (e.g. collection/s products) or upward (e.g. bundles containing a product).
+    
+    Property values are cast to string in responses due to limitations of JSON typing, and should be interpreted by the client/user according to the data dictionary.
```
tloubrieu-jpl, line 5: I feel like we should be more explicit with the null case because it is not documented in a data dictionnary. You could keep the sentence and add an example with "null" for the field that Anil mentionned.

The integration test failing are due to another PR from Al not merged yet. At least this is what I am assuming.

